PR #4, opened 2022-02-13
Feature/nu error log

Niuslar:
Add logger using UART


Lib/Inc/error_log.h
```diff
@@ -0,0 +1,40 @@
...
+ *      Author: niuslar
+ */
+
+#ifndef ERROR_LOG_H_
+#define ERROR_LOG_H_
+
+#include <string>
+#include "uart_com.h"
+
+class CLog
```
Vavat: Is this a logging class or error logging class? Perhaps more verbose name would help. Generally, the rule is that class name and file name should match.

Lib/Inc/error_log.h
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
```
Vavat: ctor and dtor should be first methods.
``` 
class CClassName
{
  CClassName(...); //ctor - always comes first unless it's special cases like Singleton class
  ~CClassName(...); //dtor - always comes after ctor.
// public methods - these present the most interest to the user of the class, so they should be high up.

// protected methods - if they exist

// private methods - may not even declare them in a .h file, but people usually do.

// public members - if they exist they are secodn most important information for the user.

// protected members - if they exist

// private members - at the bottom because least interesting to the user.
```

Lib/Src/cpp_link.cpp
```diff
@@ -16,6 +17,16 @@ extern "C"
 
     void cpp_main()
     {
+        CUartCom uart_for_errors(huart2);
+
+        CLog log_main(&uart_for_errors, "Main");
+        log_main.setLogLevel(CLog::LOG_WARNING);
+
+#ifdef DEBUG
+        log_main.log(CLog::LOG_INFO, "Entered cpp_main function");
```
Vavat, line 26: That's nice. I like it.

Lib/Src/error_log.cpp
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+#include "error_log.h"
+
+/**
+ * @brief Constructor
+ * @param Pointer to CUartCom object
+ * @param module_name Identifier that will be added to the message
+ */
+CLog::CLog(CUartCom *error_uart, const std::string module_name)
```
Vavat: error_uart is a pointer, so suggest using p_error_uart.

Lib/Src/error_log.cpp
```diff
@@ -0,0 +1,67 @@
...
+    : mp_error_uart(error_uart),
+      m_module_name(module_name)
+{
+}
+
+/**
+ * @brief change log level to change what is sent
```
Vavat: Worth explaining what happens to messages that are below log level. Are they lost or are they stored?

Lib/Src/error_log.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * @param log_level:
+ * 			LOG_ERROR = will show only errors
+ * 			LOG_WARNING = will show error and warnings
+ * 			LOG_INFO = will show everything
+ */
+void CLog::setLogLevel(log_level_t log_level)
+{
+    m_log_level = log_level;
```
Vavat: Can someone force log_level to be a value outside enum type?

Lib/Src/error_log.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * @brief method to log errors, warnings and info
+ * @param log_level Level of the message (LOG_ERROR, LOG_WARNING, LOG_INFO)
+ * @param message
+ */
+void CLog::log(log_level_t log_level, const std::string &message)
+{
+    if (log_level == LOG_INFO && m_log_level >= LOG_INFO)
+    {
+        // Convert to C style string
+        const char *msg = ((const std::string)("[INFO]->" + m_module_name +
```
Vavat: Where is the *msg pointing? Who owns that memory? When `mp_erorr_uart->sendMessage(msg) is executed, does it copy the message or is it going to just use the pointer as if memory is safe?

Lib/Src/error_log.cpp
Niuslar: Added a case to the log() method in case someone does that.

Lib/Src/error_log.cpp
Niuslar: Thanks, I realised c_str() wasn't very reliable. I changed this method to use C++ strings instead and moved the conversion to the CUartCom sendMessage method.

Niuslar:
I think I addressed all your comments. Thanks for the input.

PR #8, opened 2022-02-14
Feature/nu adc

Niuslar:
ADC Class that will be the base to read temperature and telemetry sensors.

### DIFF
--- a/Lib/Inc/adc_data.h
+++ b/Lib/Inc/adc_data.h
@@ -20,6 +20,7 @@ public:
     ~CAdc();
 
     // Public methods
+    // use [] operator to read a particular ADC Channel
     uint16_t operator[](uint8_t adc_channel);
     void triggerAdc();
 
--- a/Lib/Inc/adc_data.h
+++ b/Lib/Inc/adc_data.h
@@ -1,0 +1,32 @@
+/**
+ * @file adc_data.h
+ */
+
+/*
+ *  Created on: Feb 14, 2022
+ *      Author: niuslar
+ */
+
+#ifndef ADC_DATA_H_
+#define ADC_DATA_H_
+
+#include "adc.h"
+#include "log.h"
+
+class CAdc
+{
+public:
+    CAdc(ADC_HandleTypeDef *p_hadc, uint8_t adc_channels);
+    ~CAdc();
+
+    // Public methods
+    uint16_t readChannel(uint8_t adc_channel);
+    void triggerAdc();
+
+private:
+    ADC_HandleTypeDef *mp_hadc;
+    uint8_t m_adc_channels;
+    uint16_t *mp_adc_data_buf;
+};
+
+#endif /* ADC_DATA_H_ */
--- a/Lib/Inc/adc_data.h
+++ b/Lib/Inc/adc_data.h
@@ -20,7 +20,7 @@ public:
     ~CAdc();
 
     // Public methods
-    uint16_t readChannel(uint8_t adc_channel);
+    uint16_t operator[](uint8_t adc_channel);
     void triggerAdc();
 
 private:
--- a/Lib/Inc/adc_data.h
+++ b/Lib/Inc/adc_data.h
@@ -28,7 +28,7 @@ public:
     void init();
     // use [] operator to read a particular ADC Channel
     uint16_t operator[](uint8_t adc_channel);
-    void triggerRead();
+    void trigger();
 
 private:
     ADC_HandleTypeDef *const mp_hadc;
--- a/Lib/Inc/adc_data.h
+++ b/Lib/Inc/adc_data.h
@@ -31,8 +31,8 @@ public:
     void triggerRead();
 
 private:
-    ADC_HandleTypeDef *mp_hadc;
-    uint8_t m_adc_channels = ADC_CHANNELS;
+    ADC_HandleTypeDef *const mp_hadc;
+    const uint8_t m_adc_channels = ADC_CHANNELS;
     uint16_t m_adc_data_buf[ADC_CHANNELS];
 };
 
--- a/Lib/Inc/adc_data.h
+++ b/Lib/Inc/adc_data.h
@@ -13,21 +13,27 @@
 #include "adc.h"
 #include "log.h"
 
-class CAdc
+/**
+ *  @brief Number of active ADC Channels,
+ *   change this according to hardware
+ */
+#define ADC_CHANNELS 14
+
+class CAdcData
 {
 public:
-    CAdc(ADC_HandleTypeDef *p_hadc, uint8_t adc_channels);
-    ~CAdc();
+    CAdcData(ADC_HandleTypeDef *p_hadc);
 
     // Public methods
+    void init();
     // use [] operator to read a particular ADC Channel
     uint16_t operator[](uint8_t adc_channel);
-    void triggerAdc();
+    void triggerRead();
 
 private:
     ADC_HandleTypeDef *mp_hadc;
-    uint8_t m_adc_channels;
-    uint16_t *mp_adc_data_buf;
+    uint8_t m_adc_channels = ADC_CHANNELS;
+    uint16_t m_adc_data_buf[ADC_CHANNELS];
 };
 
 #endif /* ADC_DATA_H_ */
--- a/Lib/Src/adc_data.cpp
+++ b/Lib/Src/adc_data.cpp
@@ -75,7 +75,7 @@ CAdc::~CAdc()
  * @brief read data from a specific adc_channel
  * @param adc_channel Number of channel starting from 0
  */
-uint16_t CAdc::readChannel(uint8_t adc_channel)
+uint16_t CAdc::operator[](uint8_t adc_channel)
 {
     // Check channel is within defined range
     if (adc_channel >= m_adc_channels)
@@ -91,6 +91,10 @@ uint16_t CAdc::readChannel(uint8_t adc_channel)
     return mp_adc_data_buf[adc_channel];
 }
 
+/**
+ * @brief Trigger a sequence read of all ADC Channels
+ * @note This function will replace all the values in the adc_data_buf.
+ */
 void CAdc::triggerAdc()
 {
     // Change ADCSTART bit in the ADC control register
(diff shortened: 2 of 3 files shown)


Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+    // Check mp_hadc is not null
+    if (!mp_hadc)
+    {
+        // Send error message
+        log_adc.log(CLog::LOG_ERROR, "Invalid pointer to ADC Handler");
+    }
+
+    // Allocate memory to store the incoming data
+    mp_adc_data_buf = new uint16_t[adc_channels];
```
Vavat: ouff. this is dangerous. runtime memory allocation error is very hard to catch. I suggest you statically allocate space for this data. 
You could do a fancy compile time gates.
```
// comment out if you want memory to be statically allocated.
#define DYNAMIC_MEMORY
#ifdef DYNAMIC_MEMORY
    mp_adc_data_buf = new uint16_t[adc_channels];
#elif
    mp_adc_data_buf = p_statically_allocated_memory;
#endif
```

Similarly you allocate the memory conditionally in .h file depending on the `#define` flag.

Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    // Calibrate ADC
+    if (HAL_ADCEx_Calibration_Start(p_hadc, ADC_SINGLE_ENDED) != HAL_OK)
```
Vavat: This will fail. Constructors are executed when classes are created, which will happen before anything else is executed. You constructor should do basic checks and that's it. No calls to hardware at all. Then you create a special `init()` method that you run to do all of this fancy stuff.

Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+    {
+        // Free memory and log error
+        delete[] mp_adc_data_buf;
+        log_adc.log(CLog::LOG_ERROR, "ADC Calibration failed");
+    }
+
+    // Start ADC with DMA
+    if (HAL_ADC_Start_DMA(p_hadc, (uint32_t *)mp_adc_data_buf, adc_channels) !=
```
Vavat: Same here.

Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+        // Free memory and log error
+        delete[] mp_adc_data_buf;
+        log_adc.log(CLog::LOG_ERROR, "ADC Start DMA failed");
+    }
+}
+
+CAdc::~CAdc()
+{
+    // Free memory when the object is destroyed
+    delete[] mp_adc_data_buf;
```
Vavat: What happens if this class failed to create and memory has not been allocated? Should you check if pointer is valid before deleting it?

Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/**
+ * @brief read data from a specific adc_channel
+ * @param adc_channel Number of channel starting from 0
+ */
+uint16_t CAdc::operator[](uint8_t adc_channel)
```
Vavat: This is really cool. Very elegant.

Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return mp_adc_data_buf[adc_channel];
+}
+
+/**
+ * @brief Trigger a sequence read of all ADC Channels
+ * @note This function will replace all the values in the adc_data_buf.
+ */
+void CAdc::triggerAdc()
```
Vavat: With c++ because the method belongs to a class you don't need to call it `triggerAdc()`. You can simply call it `trigger()` because when you use it it will look like:
```
CAdc temperature_sensors(&hadc1, 8);
...
temperature_sensors.trigger();
```

Lib/Src/cpp_link.cpp
```diff
@@ -15,24 +17,46 @@ extern "C"
...
+#define ADC_CHANNELS 14
+
     void cpp_main()
     {
         CUartCom uart_for_errors(&huart2);
 
         CLog log_main(&uart_for_errors, "Main");
         log_main.setLogLevel(CLog::LOG_INFO);
 
+        CAdc adc_1(&hadc, ADC_CHANNELS);
```
Vavat: Right, so this works because you are creating the class within cpp_main(). If someone was to use your class and create it as a global variable, it'd fail. Test it and see.

Lib/Src/cpp_link.cpp
```diff
@@ -15,24 +17,46 @@ extern "C"
...
-        // C++ Code here
 
         // Infinite Loop
         while (1)
         {
+#ifdef DEBUG
+            // Test 3 of the 14 ADC channels
+
+            // First read channels
+            uint16_t adc_ch_1 = adc_1[0];
```
Vavat, line 41: This is excellent. However... it is obscure and might be confusing to a noob. But... it is very elegant. I like it.
Not something you need to change, but think about your choices here between being elegant and writing maintainable code. Maintainable by anyone.

Vavat: To clarify, are you happy to sacrifice accessibility to the code to some people because they are very new to c++ and not as smart or experienced for sake of beautiful code. This decision you should be very wary of depending on the team you are working in.

Lib/Src/cpp_link.cpp
```diff
@@ -15,24 +17,46 @@ extern "C"
 {
 #endif
 
+#define ADC_CHANNELS 14
+
```
Vavat: Try creating your CAdc instance here.

Lib/Src/adc_data.cpp
Vavat: Also, make sure your constructor leaves the class in a safe state in case init() is never executed.

Lib/Src/adc_data.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * @brief Calibrate and Start ADC with DMA
+ * @note The ADC will read all the channels once and then stop.
+ * To read the channels again, use the triggerADC() method.
+ * @param p_hadc Pointer to ADC handler
+ * @param adc_channels Number of active channels for the ADC
+ * 		(This depends on the hardware configuration)
+ */
+CAdc::CAdc(ADC_HandleTypeDef *p_hadc, uint8_t adc_channels)
```
Vavat: Class name does not match file name.

Lib/Inc/adc_data.h
```diff
@@ -0,0 +1,33 @@
...
+    ~CAdc();
+
+    // Public methods
+    // use [] operator to read a particular ADC Channel
+    uint16_t operator[](uint8_t adc_channel);
+    void triggerAdc();
+
+private:
+    ADC_HandleTypeDef *mp_hadc;
+    uint8_t m_adc_channels;
```
Vavat: this should be `const` to make it safer against someone modifying your code later.

Lib/Inc/adc_data.h
```diff
@@ -0,0 +1,33 @@
...
+    uint16_t *mp_adc_data_buf;
```
Vavat: This should be `uint16_t * const` to make it safer.

Lib/Inc/adc_data.h
```diff
@@ -0,0 +1,33 @@
...
+    CAdc(ADC_HandleTypeDef *p_hadc, uint8_t adc_channels);
+    ~CAdc();
+
+    // Public methods
+    // use [] operator to read a particular ADC Channel
+    uint16_t operator[](uint8_t adc_channel);
+    void triggerAdc();
+
+private:
+    ADC_HandleTypeDef *mp_hadc;
```
Vavat: Should probably make this `* const` too.

Vavat (commented):
Some clean up is recommended. Mostly style changes.

Lib/Src/adc_data.cpp
Niuslar: I think using a ```#define ADC_CHANNELS 14``` in the header file is the best option. Easy to change and less prone to error.

Lib/Src/cpp_link.cpp
Niuslar: I think I'll leave it in this case. It makes it clearer that the channels start from 0 and it is easy to use. And if someone new to C++ doesn't know about operator overloading it could be a good learning opportunity.

Lib/Inc/adc_data.h
```diff
@@ -13,21 +13,27 @@
...
-    CAdc(ADC_HandleTypeDef *p_hadc, uint8_t adc_channels);
-    ~CAdc();
+    CAdcData(ADC_HandleTypeDef *p_hadc);
 
     // Public methods
+    void init();
     // use [] operator to read a particular ADC Channel
     uint16_t operator[](uint8_t adc_channel);
-    void triggerAdc();
+    void triggerRead();
```
Vavat: meh. I am still not liking the name. Simle just looks nicer.
```
adc_data.trigger();
```
But this is just a matter of taste in this case. You generally make decent choices as far as names go.

Lib/Inc/adc_data.h
```diff
@@ -13,21 +13,27 @@
 #include "adc.h"
 #include "log.h"
 
-class CAdc
+/**
+ *  @brief Number of active ADC Channels,
+ *   change this according to hardware
+ */
+#define ADC_CHANNELS 14
```
Vavat, line 20: What you could do is allocate 20 slots here, but allow for ctor to have less channels used. It'll be wasted memory, but only very little. To be considered.

Lib/Src/cpp_link.cpp
Vavat: I agree with your decision. The point I was trying to make is that at some stage in your career, you'll be in charge of a team. This is when you need to learn to accomplish things using someone else's skillset. If you have mixed skill team, it is useful to keep in mind that higher skill people might not be writing code with lower skill colleagues in mind. As a leader it'll be your job to make sure everyone is comfortable. You'll recognise what I am talking about when you see it in reality.

Vavat (approved):
Other than the name choice, this is good.

PR #9, opened 2022-02-15
Add getVolts() method

### LINKED ISSUE #10: Create getVolts method
The CAdcData class is incomplete and needs a method that returns the ADC readings converted to volts. 
Add getVolts method to fix this

Niuslar:
Fix #10

### DIFF
--- a/Lib/Src/adc_data.cpp
+++ b/Lib/Src/adc_data.cpp
@@ -11,7 +11,8 @@
 
 #include "adc_data.h"
 
-#define ADC_START_BIT 2
+#define ADC_START_BIT        2
+#define ADC_CONVERSION_COEFF ((float)ADC_VDDA / ADC_RES)
 
 CUartCom uart_adc(&huart2);
 
@@ -54,10 +55,11 @@ void CAdcData::init()
 }
 
 /**
- * @brief read data from a specific adc_channel
+ * @brief read data converted to volts for a specific adc_channel
  * @param adc_channel Number of channel starting from 0
+ * @retval Volts
  */
-uint16_t CAdcData::operator[](uint8_t adc_channel)
+float CAdcData::operator[](uint8_t adc_channel)
 {
     // Check channel is within defined range
     if (adc_channel >= m_adc_channels)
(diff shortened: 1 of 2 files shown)


Niuslar:
@Vavat Added a getVolts method to use for temperature and telemetry.

Lib/Src/adc_data.cpp
```diff
@@ -73,6 +73,15 @@ uint16_t CAdcData::operator[](uint8_t adc_channel)
     return m_adc_data_buf[adc_channel];
 }
 
+float CAdcData::getVolts(uint8_t adc_channel)
+{
+    // The [] operator checks the channel and returns 0 in case of error
+    uint16_t raw_data = this->operator[](adc_channel);
```
Vavat: you can just access the array. will be a lot clearer.
```
uint16_t raw_data = m_adc_data_buf[adc_channel];
```

Lib/Src/adc_data.cpp
```diff
@@ -73,6 +73,15 @@ uint16_t CAdcData::operator[](uint8_t adc_channel)
...
+    float volts = ((float)raw_data) / (ADC_RES)*ADC_VDDA;
```
Vavat: (ADC_RES)*(ADC_VDDA) are compile time known values, so you should precalculate them into some sort of coefficient and then only use multiply because it's faster than divide.
```
#define ADC_CONV_COEFF (1/(ADC_RES*ADC_VDDA))
...
 float volts = ((float)raw_data) * ADC_CONV_COEFF;
```

Vavat (commented):
The PR is confusing because there is no description of what you are doing here. If you had tickets, then it would be sufficient to just mention ticket number. Without tickets it's hard for me to know what you are trying to achieve and therefore hard to know whether I think you were successful.

Lib/Src/adc_data.cpp
Niuslar: I could, but then I would need to repeat the code to check the channel is valid. Maybe if I change the operator to a function is clearer? ```uint16_t raw_data = this->getData(adc_channel)```.

Niuslar:
You're right. Added a ticket to make things clearer. Thanks.

Lib/Src/adc_data.cpp
Vavat: Do you need to obtain raw ADC counts? When would it ever be useful? I think your operator should return voltage and not counts. This way you keep the elegant solution and remove unnecessary methods.

Niuslar: Fixed, now the operator[] returns the voltage.

Vavat (approved):

PR #12, opened 2022-02-15
Feature/nu thermistor

### LINKED ISSUE #11: Temperature Sensor Class
Temperature Sensor Class needed to return ADC data converted to Celsius. This will be key for the PID Control. 
~~The class should also have a method to send the Temperature reading via UART.~~ Not sure this is a good idea anymore. I could use the CLog class to do this.

Niuslar:
Simple Thermistor class with a method to get the converted temperature for the selected ADC channel. 
I made it so you can have one object per ADC Channel, I think this makes it easier to use in case the ADC channels are not in order.


Lib/Inc/thermistor.h
```diff
@@ -0,0 +1,28 @@
...
+ *  Created on: Feb 15, 2022
+ *      Author: niuslar
+ */
+
+#ifndef THERMISTOR_H_
+#define THERMISTOR_H_
+
+#include "adc_data.h"
+
+class CThermistor
```
Vavat: Filename mismatch.

Niuslar: Fixed

Lib/Inc/CThermistor.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "adc_data.h"
+
+class CThermistor
+{
+public:
+    CThermistor(CAdcData &adc_data, uint8_t adc_channel);
+
+    // Public methods
+    float getTemperature();
```
Vavat: This is somewhat weird, but it would be better if thermistor was a standalone class that took voltage and produced temperature. The reasons for it are:
1. Code reuse becomes easier if your class is not attached to an ADC class.
2. It is easiesr to test your code if you can feed it artificial voltage values for which you know what temperature should be returned. It allows for static testing of the code.
So I recommend this class is redone as follows:
```
CThermistor::CThermistor(float* p_calibration_array, uint8_t calibration_order)
{
    // calibration array contains a number of floats that participate in calibration of the thermistor
    // calibration order tells constructor how many variables are in the calibration array.
}

File: Lib/Inc/CThermistor.h [hunk 3]
```diff
@@ -0,0 +1,28 @@
...
+
+private:
+    CAdcData &m_adc_data;
```
Vavat: Here you need to create an array of calibration values in float.

Lib/Src/CThermistor.cpp
```diff
@@ -0,0 +1,59 @@
...
+/*
+ *  Created on: Feb 15, 2022
+ *      Author: niuslar
+ */
+
+#include "CThermistor.h"
+#include "math.h"
+
+#define MIN_VOLT_RANGE     (0.57)
+#define MAX_VOLT_RANGE     (2.67)
```
Vavat: Min and Max values should probably be a setting that you can give to the thermistor either as part of constructor or as a separate method. Other thermistors will have different limits, so no need to hardcode the values.

Lib/Src/CThermistor.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+    // Check channel
+    if (m_adc_channel >= ADC_CHANNELS)
+    {
+        // Log warning
+        log_temp.log(CLog::LOG_WARNING, "Invalid ADC Channel");
+    }
+}
+
+float CThermistor::getTemperature()
```
Vavat:
```
float CThermistor::getTemperature(float voltage)
{
    float temperature = 0;
    // here you multiply voltage by the calibration coefficients stored in the thermistor.
    return temperature;
}

File: Lib/Src/cpp_link.cpp [hunk 6]
```diff
@@ -42,15 +48,20 @@ extern "C"
             uint16_t adc_ch_2 = adc_1[1];
             uint16_t adc_ch_14 = adc_1[13];
 
+            float temp = thermistor_1.getTemperature();
```
Vavat:
```
float temp_1 = thermistor.getTemperature(adc_1[1]);
```

Vavat (commented):
The CThermistor should only be handling thermistor. It does not need to know about ADC or channels or anything else that is not directly thermistor related. This class converts voltage into temperature and nothing else. Pure and simple class.

Lib/Src/CThermistor.cpp
Niuslar: Do you mean something like: 
```
 temp_celsius = m_coeff_array[0] - m_coeff_array[1] * voltage + m_coeff_array[2] * pow(voltage, 2) - m_coeff_array[3] * pow(voltage, 3);
```
?

Lib/Inc/CThermistor.h
```diff
@@ -0,0 +1,34 @@
...
+
+/*
+ *  Created on: Feb 15, 2022
+ *      Author: niuslar
+ */
+
+#ifndef CTHERMISTOR_H_
+#define CTHERMISTOR_H_
+
+#include "CAdcData.h"
```
Vavat: Not necessary.

Lib/Src/CThermistor.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+/*
+ *  Created on: Feb 15, 2022
+ *      Author: niuslar
+ */
+
+#include "CThermistor.h"
+#include "math.h"
+
+#define OUT_OF_RANGE_VALUE (999.9)
```
Vavat: You should make this 
```
public: 
static float const CThermistor::OUT_OF_RANGE = 999.9;
```
That way it can be referenced outside the class.
```
if(thermistor.getTemperature(voltage) == CThermistor::OUT_OF_RANGE)
{
    log.send("sensor fault");
}
```

Lib/Src/CThermistor.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+CUartCom uart_temp(&huart2);
+CLog log_temp(&uart_temp, "Temp_sensor");
```
Vavat: Would it be possible to create the logger outside this class and then pass the pointer during construction? Make two constructors with and without logger. This way class can function even without logger.

Lib/Src/CThermistor.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * @param min_voltage Minimum voltage limit
+ * @param max_voltage Max voltage limit
+ */
+void CThermistor::setLimits(const float min_voltage, const float max_voltage)
+{
+    m_min_volt_limit = min_voltage;
+    m_max_volt_limit = max_voltage;
+}
+
+float CThermistor::getTemperature(float voltage)
```
Vavat: You can declare this method `const`. This way it can be optimised further by the compiler. Have a look what that means. Can also look at what `pure` method does.

Vavat: ah. that might not work because you are calling a logger method, which is not const.

Vavat (dismissed):
Some minor changes recommended, but otherwise looks good.

Lib/Src/CThermistor.cpp
Vavat: ah. I forgot about that.
Nope. Let me do a branch and show you.

Lib/Src/CThermistor.cpp
Niuslar: Can't I add it without the `CThermistor::` ?
```
public:
static float const OUT_OF_RANGE = 999.9;
```

Niuslar:
I added the option to use the constructor with and without calibration parameters.

Lib/Inc/CThermistor.h
```diff
@@ -0,0 +1,50 @@
...
+                const float *p_calibration_coeff = s_calibration_coeff,
+                const uint8_t calibration_order = s_calibration_order);
+
+    // Public methods
+    float getTemperature(float voltage) const;
+    void setLimits(float min_voltage, float max_voltage);
+
+    // Public variables
+    static constexpr float OUT_OF_RANGE = 999.9;
+    static const float s_calibration_coeff[];
```
Vavat: MIght be better to make this variable more distinct than just prefix. 
`s_calibration_coeff[]` and `m_calibration_coeff[]` are very similar and might cause confusion in the future. Say:
`s_default_coeff[]` and same for `s_default_order`.

Lib/Inc/CThermistor.h
```diff
@@ -0,0 +1,50 @@
...
+    static const uint8_t s_calibration_order;
+
+private:
+    float m_min_volt_limit = MIN_VOLT_RANGE;
+    float m_max_volt_limit = MAX_VOLT_RANGE;
+    const float *mp_calibration_coeff;
```
Vavat: I think this is meant to be an array of MAX_ORDER+1 to store coefficients. We also need a method to set the coefficients in case we want to recalibrate the thermistors by external command.

Vavat: And it cannot be `const`. this must be remnants from when I wrote it carelessly.

Lib/Inc/CThermistor.h
Vavat: You could look into using std::vector for this purpose actually.

Lib/Inc/CThermistor.h
Vavat: You could look into using `std::vector` for this purpose.

Lib/Inc/CThermistor.h
```diff
@@ -0,0 +1,50 @@
...
+    const uint8_t m_calibration_order;
```
Vavat: this should not be `const` either if we allow recalibration.

Lib/Inc/CThermistor.h
Niuslar: I don't understand why this needs to be an array? I saw in your code that you copied each element, but why can't I store a pointer to the original array? 
You're right about the const. I meant to make the pointer const `float * const mp_calibration_coeff;`.

Lib/Src/CThermistor.cpp
```diff
@@ -0,0 +1,122 @@
...
+ * @param calibration_order Order of the polynomial equation
+ *
+ * @note The order of the polynomial is used in a strict mathematical sense,
+ i.e. third order polynomial will have 4 coefficients: a0, a1, a2, a3. The
+ minimum number of coefficients required is 2, which would make for a first
+ order polynomial otherwise knows as linear relation.
+ */
+CThermistor::CThermistor(const float *p_calibration_coeff,
+                         const uint8_t calibration_order)
+    : mp_calibration_coeff(p_calibration_coeff),
```
Vavat: Should copy contents into local member variable. All of the coefficients. we want for them to be local.

Lib/Inc/CThermistor.h
```diff
@@ -0,0 +1,50 @@
...
+
+    // Constructor with logger
+    CThermistor(CLog *p_logger,
+                const float *p_calibration_coeff = s_calibration_coeff,
+                const uint8_t calibration_order = s_calibration_order);
+
+    // Public methods
+    float getTemperature(float voltage) const;
+    void setLimits(float min_voltage, float max_voltage);
+
```
Vavat, line 32:
```
void setCalibration(float *p_calibration_coeff, uint8_t calibration_order)
{
    // copy the contents of the pointer up to calibration_order+1 into member variable.
}

Vavat (commented):
I think I caused some confusion by being careless with my example. Have a look at comments and let me know what you think. 
Also, decision to keep arguments as `const` is probably useless. Might keep is, but if it gets in the way, remove it. It's borderline paranoia.

Niuslar:
@Vavat I merged the changes we talked about earlier. I had to change the constructor a bit to make things work.  
I also added a definition for the setCalibration method, but I left a TODO that needs to be addressed. 

I also removed the logger, I agree that is better to just make it part of the controller.

Vavat (approved):

PR #13, opened 2022-02-15
Feature/srm gpio wrapper class

Vavat:
Create a generic wrapper for the GPIO pins. This is meant in the future to support EXTI and change callbacks.

### DIFF
--- a/.cproject
+++ b/.cproject
@@ -41,7 +41,7 @@
 								</option>
 								<option IS_BUILTIN_EMPTY="false" IS_VALUE_EMPTY="false" id="com.st.stm32cube.ide.mcu.gnu.managedbuild.tool.c.compiler.option.includepaths.1721070164" name="Include paths (-I)" superClass="com.st.stm32cube.ide.mcu.gnu.managedbuild.tool.c.compiler.option.includepaths" useByScannerDiscovery="false" valueType="includePath">
 									<listOptionValue builtIn="false" value="../Core/Inc"/>
-									<listOptionValue builtIn="false" value="&quot;${workspace_loc:/${ProjName}/Lib/Inc}&quot;"/>
+									<listOptionValue builtIn="false" value="../Lib/Inc"/>
 									<listOptionValue builtIn="false" value="../Drivers/STM32L0xx_HAL_Driver/Inc"/>
 									<listOptionValue builtIn="false" value="../Drivers/STM32L0xx_HAL_Driver/Inc/Legacy"/>
 									<listOptionValue builtIn="false" value="../Drivers/CMSIS/Device/ST/STM32L0xx/Include"/>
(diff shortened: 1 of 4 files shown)


Lib/Inc/CGpioWrapper.h
```diff
@@ -0,0 +1,28 @@
+/*
```
Niuslar: I think this should follow the doxygen style for the file to be included?
```   
/**
/* @file CGpioWrapper.h
/*
```

Lib/Inc/CGpioWrapper.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef CGPIOWRAPPER_H_
+#define CGPIOWRAPPER_H_
+
+#include "gpio.h"
+
+class CGpioWrapper
+{
+public:
+    CGpioWrapper(GPIO_TypeDef *p_port, uint32_t pin);
```
Niuslar: If you're using the GPIO_PIN_X format, the max value is 0xFFFF (all pins), so no need for 32 bits. ```uint16_t``` should be enough.

Lib/Src/CGpioWrapper.cpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+CGpioWrapper::CGpioWrapper(GPIO_TypeDef *p_port, uint32_t pin)
+    : mp_port(p_port),
+      m_pin(pin)
+{
+    if (mp_port == NULL)
+    {
+        Error_Handler();
+    }
+    if (!IS_GPIO_PIN(m_pin))
```
Niuslar: If you use ```uint16_t``` I think this becomes redundant. Here's the IS_GPIO_PIN macro, let me know if you agree. 
```
#define GPIO_PIN_MASK              (0x0000FFFFU) /* PIN mask for assert test */
#define IS_GPIO_PIN(__PIN__)        ((((uint32_t)(__PIN__) & GPIO_PIN_MASK) != 0x00U) &&\
                                     (((uint32_t)(__PIN__) & ~GPIO_PIN_MASK) == 0x00U))
```

Niuslar (approved):
Some very minor comments.

Lib/Inc/CGpioWrapper.h
Vavat: But the CubeMX defines pins as uint32_t, so if I use uint16_t the compiler is going to give warnings about mismatched variables and reduced scope.

Lib/Inc/CGpioWrapper.h
Vavat: I did not know that. Cool.

Lib/Src/CGpioWrapper.cpp
Niuslar: Just realised it's still useful to check the pin is not 0x00.

Lib/Inc/CGpioWrapper.h
Vavat: Turns out I was wrong that CubeMX defines them as 32 bit unsigned int. they are just #define values. I will change it.

Lib/Src/CGpioWrapper.cpp
Vavat: It also checks that only one bit is set in a mask.

.cproject
```diff
@@ -237,4 +238,5 @@
 		</scannerConfigBuildInfo>
 	</storageModule>
 	<storageModule moduleId="org.eclipse.cdt.make.core.buildtargets"/>
+	<storageModule moduleId="refreshScope"/>
```
Niuslar, line 241: I think our .cproject files will always be different, perhaps it is better to ignore them and add a "settings setup" section in the ReadMe. (Add Lib to the include path, untick exclude from build, add support for floating point in printf, etc).

.project
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>Multichannel temperature sensor CPP</name>
+	<name>Multichannel temperature sensor</name>
```
Niuslar, line 3: I think this one should be safe to ignore as well.

.cproject
Vavat: Don't know. Perhaps worth researching what .cproject file does.

Niuslar: .project and .cproject are the IDE configuration files. .cproject is specific to the CDT (C/C++ plugin) and according to t[his link](https://www.eclipse.org/forums/index.php/t/174633/) 
> This defines things that are specific to C/C++ projects such as include
paths, library paths, and build commands.

PR #16, opened 2022-02-16
Update ReadMe

Niuslar:
I added a **Project Setup** section. Even if we keep the .cproject file, I think this could be useful.


Vavat (approved):
Agree. This is good information to have either way.
Don't remove .cproject just yet. Let's research how others do it. Might want to ask on reddit.

PR #18, opened 2022-02-18
Feature/srm basic super loop structure

Vavat:
Basic superloop structure that involves three classes.
IComChannel - base abstract class for communication channel interface. Establishes the API to communication channels. All communication channel classes must inherit form IComChannel.
CController - base abstract class for running controllers. Establishes the API to controllers. All controllers that must run regularly must inherit from CController.
CDispatcher - main superloop execution class.

### DIFF
--- a/Lib/Inc/CDispatcher.h
+++ b/Lib/Inc/CDispatcher.h
@@ -26,11 +26,15 @@ public:
 
     bool registerController(CController *p_controller);
     bool registerComChannel(IComChannel *p_comchannel);
+    void run();  // TODO: figure out a reliable way to declare this as noreturn.
 
 private:
+    void processComChannels();
+
     CLog *mp_logger;
     CController *mp_controllers[MAX_CONTROLLERS];
     uint8_t m_controller_count;
+    uint8_t m_active_controller;
 
     IComChannel *mp_comchannels[MAX_COMCHANNELS];
     uint8_t m_comchannel_count;
--- a/Lib/Inc/CDispatcher.h
+++ b/Lib/Inc/CDispatcher.h
@@ -9,6 +9,7 @@
 #define CDISPATCHER_H_
 
 #include <stdio.h>
+#include <string>
 
 #define MAX_CONTROLLERS 20
 #define MAX_COMCHANNELS 5
@@ -28,10 +29,10 @@ public:
 
 private:
     CLog *mp_logger;
-    CController *p_controllers[MAX_CONTROLLERS];
+    CController *mp_controllers[MAX_CONTROLLERS];
     uint8_t m_controller_count;
 
-    IComChannel *p_comchannels[MAX_COMCHANNELS];
+    IComChannel *mp_comchannels[MAX_COMCHANNELS];
     uint8_t m_comchannel_count;
 };
 
--- a/Lib/Inc/CDispatcher.h
+++ b/Lib/Inc/CDispatcher.h
@@ -1,0 +1,38 @@
+/*
+ * CDispatcher.h
+ *
+ *  Created on: 16 Feb 2022
+ *      Author: salavat.magazov
+ */
+
+#ifndef CDISPATCHER_H_
+#define CDISPATCHER_H_
+
+#include <stdio.h>
+
+#define MAX_CONTROLLERS 20
+#define MAX_COMCHANNELS 5
+
+#include "CController.h"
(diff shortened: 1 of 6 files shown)


Lib/Inc/CDispatcher.h
```diff
@@ -0,0 +1,43 @@
+/*
```
Niuslar, line 6: Missing doxygen style
```
/**
/* @file ....
```

Lib/Inc/IComChannel.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * IComChannel.h
+ *
+ *  Created on: 16 Feb 2022
+ *      Author: salavat.magazov
+ */
+
+#ifndef INC_ICOMCHANNEL_H_
```
Niuslar, line 8: name missmatch

Lib/Src/CDispatcher.cpp
```diff
@@ -0,0 +1,152 @@
+/*
```
Niuslar, line 6: same here about the doxygen style

Lib/Src/CDispatcher.cpp
```diff
@@ -0,0 +1,207 @@
...
+ * CDispatcher.cpp
+ *
+ *  Created on: 16 Feb 2022
+ *      Author: salavat.magazov
+ */
+
+#include "CDispatcher.h"
+#include "stm32l0xx_hal.h" /* makes HAL function calls available. */
```
Niuslar, line 14: I'm not sure this `#include` is *strictly* necessary because CDispatcher.h includes CLog.h, which indirectly includes stm32l0xx_hal.h.

Lib/Src/CDispatcher.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+    // TODO Auto-generated destructor stub
+}
+
+bool CDispatcher::registerController(CController *p_controller)
+{
+    bool success = false;
+    if (p_controller == nullptr)
+    {
+        mp_logger->log(CLog::LOG_ERROR, "Invalid controller pointer.");
```
Niuslar, line 45: maybe there should be a check to see if mp_logger is nullptr before calling the log method.

Lib/Src/CDispatcher.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+        // TODO: consider having a callback within the CDispatcher class that
+        // gets registered with the IComChannel class and is called when new
+        // command arrives. This might be a better option for a more real-time
+        // response to commands.
+        processComChannels();
+    }
+}
+
+bool CDispatcher::newCommand(std::string command, IComChannel *p_comchannel)
```
Niuslar, line 142: A `@brief` for this method would be great.

Niuslar (commented):
I finally got time to review this properly and it's great. Thanks

Lib/Inc/CDispatcher.h
Vavat: fixed.

Lib/Inc/IComChannel.h
Vavat: Interesting. This was autogenerated, so not sure whey it's not matching the class name exactly. These are the include gateways though, so they are not subject to the rule about file name and class name matching. They are there to prevent recursive #include.

Lib/Src/CDispatcher.cpp
Vavat: fixed.

Lib/Src/CDispatcher.cpp
Vavat: But if someone removes the CLog for whatever reason, they'll spend ages trying to figure out why unrelated logger removal broke the timing.

Lib/Src/CDispatcher.cpp
Vavat: Agreed. I created a check in ctor. Dispatcher really should not operate without a logger.

Lib/Src/CDispatcher.cpp
Vavat: fixed doxygen on all methods.

Lib/Src/CDispatcher.cpp
Niuslar: I agree

Niuslar (approved):

PR #24, opened 2022-02-19
Feature/srm pid loop

Vavat:
Ticket #19.

### DIFF
--- a/Lib/Inc/CPIDLoop.h
+++ b/Lib/Inc/CPIDLoop.h
@@ -13,14 +13,14 @@
 #define CPIDLOOP_H_
 
 /* Default PID parameters.
- * These should be close enough to make control fast, but stable.
+ * These are setup to make PID loop transparent. Output will be equal to error.
  * Proportional is percentage of power per unit of error.
- * Integral is units of output per unit of error per minute.
- * Differential is unit of output per degree error change per minute.
+ * Integral is units of output per unit of error per unit of time.
+ * Differential is unit of output per degree error change per unit of time.
  * Normally for slow moving processes differential is zero.
  */
 #define DEFAULT_P_COEFF 1
-#define DEFAULT_I_COEFF (1.0 / (60.0 * 60.0))
+#define DEFAULT_I_COEFF 0
 #define DEFAULT_D_COEFF 0
 
 class CPIDLoop
--- a/Lib/Inc/CPIDLoop.h
+++ b/Lib/Inc/CPIDLoop.h
@@ -1,0 +1,98 @@
+/**
+ * @file CPIDLoop.h
+ */
+
+/*
+ * CPIDLoop.h
+ *
+ *  Created on: 19 Feb 2022
+ *      Author: salavat.magazov
+ */
+
+#ifndef CPIDLOOP_H_
+#define CPIDLOOP_H_
+
+/* Default PID parameters.
+ * These should be close enough to make control fast, but stable.
+ * Proportional is percentage of power per unit of error.
+ * Integral is units of output per unit of error per minute.
+ * Differential is unit of output per degree error change per minute.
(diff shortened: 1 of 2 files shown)


Vavat:
This class needs either unit testing setup or should be tested on real hardware. Real hardware seems like an easier and more robust method.

Lib/Inc/CPIDLoop.h
```diff
@@ -0,0 +1,97 @@
...
+
+    float getI() const
+    {
+        return m_i;
+    }
+    float getD() const
+    {
+        return m_d;
+    }
+    void setI(float i)
```
Niuslar, line 102: Why is `i` the only one with a `set` method?

Lib/Inc/CPIDLoop.h
```diff
@@ -0,0 +1,97 @@
...
+        float min_d;
+        float max_d;
+    };
+
+    CPIDLoop(float k_p = DEFAULT_P_COEFF,
+             float k_i = DEFAULT_I_COEFF,
+             float k_d = DEFAULT_D_COEFF);
+
+    float run(float target, float actual);
+    float run(float target, float actual, float deltaError);
```
Niuslar, line 61: I don't get the point of having 2 run methods if the first one is calling the second one to get the output. Why not have something like the following:
```
float run(float target, float actual)
{
    float error = target - actual;
    float output  = 0;
    delta_error = error - m_old_error;
    m_old_error = error;
    ... continue with the second definition of run
    return output;
}
```
?

Lib/Inc/CPIDLoop.h
```diff
@@ -0,0 +1,97 @@
...
+    float getD() const
+    {
+        return m_d;
+    }
+    void setI(float i)
+    {
+        m_i = i;
+    }
+
+    void setupLoop(const CONTROL_STRUCT_T &control_struct)
```
Niuslar: Is this the only way to change the coeff. limits? If I create a CPIDLoop instance with some kp, ki and kd values. Do I need to repeat them in the struct to change the default limits? 
```
CPIDLoop control_1(1,0,0); 
control_1.setupLoop(control_struct);  -> This control_struct needs to repeat 1,0,0?
```
Wouldn't it be easier to pass the reference to the control_struct as parameter in the CPIDLoop constructor?

Niuslar (commented):
I have some questions, but overall it looks great. Thanks

Lib/Inc/CPIDLoop.h
Vavat: Because both P and D are calculated based on instantaneous input. Integral on the other hand has "memory" of past error values. If you want to modify that "memory" you need to access the value.

Lib/Inc/CPIDLoop.h
Vavat: This one is harder to explain. It is based on reality of the measuring input signals. Sometimes the signal you are trying to control is very noisy. D-term is very sensitive to noise because it, by the mathematical relationship associated with it, amplifies the changes and noise is very fast changes. If you want to filter the signal, then it makes sense for the filter to also calculate the rate of change of the signal and feed this filtered rate to the PID loop. 
I think it would probably make more sense to you if you knew differential calculus associated with control systems. It's not something you can learn from looking at the code or a quick google look up. If you are interested I'll recommend a math channel on youtube to watch. 
The code structure essentially allows you to either feed error value (which in this case is a representation of rate of change) or allow the algorithm to calculate it from the input signal based on past values. It also avoids repetition of code.

Lib/Inc/CPIDLoop.h
Vavat: You do not need to create control_struct_t instance yourself. CPIDLoop will do that during instantiation.

Niuslar (approved):

PR #28, opened 2022-02-20
Feature/srm softpwm inherit gpio

Vavat:
This implements PWM base class, hard PWM and soft PWM.


Lib/Inc/CHardPwmOutput.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ *  Created on: 20 Feb 2022
+ *      Author: salavat.magazov
+ */
+
+#ifndef CHARDPWMOUTPUT_H_
+#define CHARDPWMOUTPUT_H_
+
+#include "IPwmOutput.h"
+#include "main.h"
```
Niuslar, line 17: Maybe it's worth including `main.h` in IPwmOutput.h? Since `main.h` includes `"stm32l0xx_hal.h"`, both, hardware and software timers would use it. I get it if you prefer to keep it explicitly in each header file.

Lib/Inc/IPwmOutput.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+/*
+ * IPwmOutput.h
+ *
+ *  Created on: 20 Feb 2022
+ *      Author: salavat.magazov
+ */
+
+#ifndef INC_IPWMOUTPUT_H_
```
Niuslar: Not that it matters, but for consistency I would change the guards created by the IDE.

Lib/Src/CHardPwmOutput.cpp
```diff
@@ -0,0 +1,97 @@
...
+ */
+void CHardPwmOutput::setDutyCycle(float duty_cycle_percent)
+{
+    if (mb_initialised)
+    {
+        m_output = duty_cycle_percent;
+        uint32_t capture_compare_reg;
+        duty_cycle_percent *= mp_timer->Instance->ARR + 1;
+        capture_compare_reg = duty_cycle_percent / 100;
+        if (capture_compare_reg > mp_timer->Instance->ARR)
```
Niuslar: This condition will be true if the duty_cycle_percent is 100. The capture_compare_reg should be compared against `mp_timer->Instance->ARR+1`.

Lib/Src/CHardPwmOutput.cpp
```diff
@@ -0,0 +1,97 @@
...
+        {
+            capture_compare_reg = mp_timer->Instance->ARR;
```
Niuslar: Same here, the max value is `mp_timer->Instance->ARR + 1`

Lib/Src/CHardPwmOutput.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * differ slightly due to discretisation within timer engine.
+ */
+void CHardPwmOutput::setDutyCycle(float duty_cycle_percent)
+{
+    if (mb_initialised)
+    {
+        m_output = duty_cycle_percent;
+        uint32_t capture_compare_reg;
+        duty_cycle_percent *= mp_timer->Instance->ARR + 1;
+        capture_compare_reg = duty_cycle_percent / 100;
```
Niuslar: Wouldn't this truncate the value instead of rounding it?

Lib/Src/CHardPwmOutput.cpp
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @return True duty cycle as generated by the timer.
+ */
+float CHardPwmOutput::getDutyCycle() const
+{
+    float duty_cycle = -1;
+    if (mb_initialised)
+    {
+        duty_cycle = __HAL_TIM_GET_COMPARE(mp_timer, m_channel);
+        duty_cycle /= mp_timer->Instance->ARR;
```
Niuslar: `mp_timer->Instance->ARR + 1`

Lib/Inc/CSoftPwmOutput.h
```diff
@@ -0,0 +1,50 @@
+/*
```
Niuslar, line 6: doxygen

Lib/Src/CSoftPwmOutput.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * method to keep the PWM operational.
+ *
+ * @param period Period of the PWM.
+ * @param p_timer Pointer to hardware timer control structure.
+ */
+void CSoftPwmOutput::init(uint32_t period, TIM_HandleTypeDef *p_timer)
+{
+    if (p_timer != nullptr)
+    {
+        HAL_TIM_RegisterCallback(sp_timer,
```
Niuslar: I think this is meant to be p_timer. `sp_timer` is `nullptr` here.

Lib/Src/CSoftPwmOutput.cpp
```diff
@@ -0,0 +1,133 @@
...
+                                 HAL_TIM_PERIOD_ELAPSED_CB_ID,
+                                 softPwmTimerTick);
+    }
+    sp_timer = p_timer;
+    s_period = period;
```
Niuslar, line 89: Shouldn't we should start the timer here? `HAL_TIM_Base_Start_IT(sp_timer)`

Lib/Src/CSoftPwmOutput.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+ * @brief Initialise the class.
+ * @note This init only needs to be called once for all instances.
+ * For operation with timer, it is assumed that timer has been fully setup.
+ * For operation without a timer, user needs to ensure repeated calls to tick()
+ * method to keep the PWM operational.
+ *
+ * @param period Period of the PWM.
```
Niuslar: Is the period in ms?

Lib/Inc/CSoftPwmOutput.h
Niuslar: and same in the src file

Lib/Inc/CHardPwmOutput.h
Vavat: The general principle is to keep scope as small as possible. Too large scope normally leads to less reusability of the code. I understand the convenience you want to pursue, but it's false convenience. It makes your life easier in this project, but when you copy this codebase to work on another project it'll be harder to adapt if your original headers permeate everywhere in libraries.

Lib/Inc/IPwmOutput.h
Vavat: I think the INC_ prefix is when you create header only, not entire class. I think it's IDE way of denoting it's an include file, not source file. We already have "I" in the name of the class that denotes that this is an interface class, so I supposed INC_ is redundant. I am undecided if I like it or not. Let me consider.

Niuslar: I think it is because it is in an INC_ is included because it is in an Inc folder. If you create a header file in the Src foulder it has SRC_ before the name.

Lib/Inc/CHardPwmOutput.h
Vavat: I've changed the direct reference to hal for a specific processor to `#include "main.h"` though. I think you were right about portability improvement there.

Lib/Inc/IPwmOutput.h
Vavat: fixed.

Lib/Inc/CSoftPwmOutput.h
Vavat: Done.

Lib/Src/CSoftPwmOutput.cpp
Vavat: No. It's in raw units. Exact timer depends on how the timer is setup. If you setup timer to update every 1ms and set period to be 100, then it's going to be 100ms. If timer updates every 2ms, then period suddenly doubles.
Perhaps worth putting that in doxygen note.

Niuslar (approved):

PR #30, opened 2022-02-21
Feature/nu uart icomchannel

Niuslar:
This is a working prototype for the new CUartCom class. It receives data using interrupts and adds it to the queue. It's not pretty and might be overcomplicated, but it works.


Lib/Inc/CUartCom.h
```diff
@@ -11,25 +11,41 @@
...
 {
-private:
-    UART_HandleTypeDef *mp_huart;
-    GPIO_TypeDef *m_uart_de_port = nullptr;
-    uint16_t m_uart_de_pin;
-
 public:
-    CUartCom(UART_HandleTypeDef *p_huart);
-    CUartCom(UART_HandleTypeDef *p_huart,
+    CUartCom(const std::string name, UART_HandleTypeDef *p_huart);
```
Vavat: The pointer to the hardware is the primary parameter, so should come first.

Lib/Inc/CUartCom.h
```diff
@@ -11,25 +11,41 @@
...
+    CUartCom(std::string name,
+             UART_HandleTypeDef *p_huart,
              GPIO_TypeDef *uart_de_port,
```
Vavat, line 29: There is a CGpioWrapper class, so you should make use of it as much as possible. Code reuse is a useful thing to do.

Lib/Inc/IComChannel.h
```diff
@@ -22,7 +22,7 @@ class IComChannel
 
     virtual bool isCommandAvailable() = 0;
     virtual std::string getCommand() = 0;
-    virtual void send(std::string message) = 0;
+    virtual void send(const std::string &message) = 0;
```
Vavat, line 30: Why is it `const` and passed by reference?

Lib/Src/CLog.cpp
```diff
@@ -47,30 +47,30 @@ void CLog::log(log_level_t log_level, const std::string &message)
             "[ERROR]->" + m_module_name + ": " + message + "\n"));
 
         // Send message using UART
-        mp_error_uart->sendMessage(full_msg);
+        mp_error_uart->send(full_msg);
```
Vavat, line 56: Again, you are modifying too many things in one PR. It gets confusing what the purpose of this branch is.

Vavat: ah... this is because the uart method changed...

Lib/Src/CAdcData.cpp
```diff
@@ -14,7 +14,7 @@
 #define ADC_START_BIT        2
 #define ADC_CONVERSION_COEFF ((float)ADC_VDDA / ADC_RES)
 
-CUartCom uart_adc(&huart2);
+CUartCom uart_adc("uart_adc", &huart2);
```
Vavat: Why does ADC need a uart?

Lib/Inc/IComChannel.h
Niuslar: Because we are not changing the string and it is passed by reference to avoid copying a string.

Niuslar:
Created a new PR for the new MCU.

PR #31, opened 2022-02-23
Change CAdcData and CThermistor design

Niuslar:
Redesigned CAdcData class to make it easier to use. Now it's an CAdcChannel class which can be used as a parent class for CThermistor and CTelemetry. 

- CAdcData is now CAdcChannel
- CAdcChannel has static methods to start the ADC and trigger readings
- CThermistor is a subclass of CAdcChannel
- CThermistor constructor has an adc_channel parameter now


Lib/Src/CAdcChannel.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+/**
+ * @brief Calibrate and Start ADC
+ * @note The ADC will read all the channels once and then stop.
+ * To read the channels again, use the triggerADC() method.
+ * @note This function needs to be called only once.
+ */
+void CAdcChannel::startADC(ADC_HandleTypeDef *p_hadc)
```
Niuslar, line 38: This is called startADC even when it's part of the CAdcChannel class because it starts the general ADC and needs to be called only once.

PR #41, opened 2022-03-28
CUartCom Class

### LINKED ISSUE #26: Adapt UART to a new base class API
Existing uart class needs to inherit from IComChannel to conform to superloop structure.

Niuslar:
@Vavat, I used some ideas from your code and others from what I already had. I still need to check the details, but it works.


Lib/Inc/IComChannel.h
```diff
@@ -27,7 +27,7 @@ class IComChannel
 
     virtual bool isCommandAvailable() = 0;
     virtual std::string getCommand() = 0;
-    virtual void send(std::string message) = 0;
+    virtual void send(const std::string &message) = 0;
```
Niuslar: I changed this to const std::string & because we are not changing the message and it is passed by reference to avoid copying it.

Lib/Src/CAdcData.cpp
```diff
@@ -14,7 +14,7 @@
 #define ADC_START_BIT        2
 #define ADC_CONVERSION_COEFF ((float)ADC_VDDA / ADC_RES)
 
-CUartCom uart_adc(&huart2);
+CUartCom uart_adc(&huart2, "ADC Data");
```
Niuslar: Once this UART branch is approved I will check if CLog is necessary. It's likely that I will delete this in the future.

Lib/Src/CCmdBuffer.cpp
```diff
@@ -0,0 +1,65 @@
+/**
+ * @file CFifoBuffer.cpp
+ */
+
+/*
+ *  Created on: Mar 28, 2022
+ *      Author: niuslar
+ */
+
+#include <CCmdBuffer.h>
```
Niuslar: Need to change this to "CCmdBuffer.h"

Niuslar:
@Vavat This is ready for review now.

Lib/Inc/CUartCom.h
```diff
@@ -11,25 +11,39 @@
...
              uint16_t uart_de_pin);
 
-    void sendMessage(const std::string &msg);
+    void startReception();
+    void send(const std::string &msg);
+    bool isCommandAvailable();
+    std::string getCommand();
+    void uartRxHandler(UART_HandleTypeDef *p_huart);
+
+    static CUartCom *sp_UART_1;
```
Vavat: Perhaps this is the place to consider having more than two instances of uart controller. Say, you have this as a static array of pointers and a static counter that counts up.

Vavat: Static array initialisation is a bit tricky. You basically need to explicitly initialise entire array to NULL.
```
CUartCom * CUartCom::sp_uart[] = {NULL, NULL, NULL...};
```

Lib/Inc/CUartCom.h
```diff
@@ -11,25 +11,39 @@
...
 public:
-    CUartCom(UART_HandleTypeDef *p_huart);
+    CUartCom(UART_HandleTypeDef *p_huart, const std::string name);
     CUartCom(UART_HandleTypeDef *p_huart,
+             const std::string name,
              GPIO_TypeDef *uart_de_port,
              uint16_t uart_de_pin);
 
-    void sendMessage(const std::string &msg);
+    void startReception();
```
Vavat: This is perhaps can be named better. The name of the method does not quite correspond with what you are doing inside. Most important thing you are doing is initialising the static variables. Fact you also enable interrupt is not really that important.

Lib/Src/CUartCom.cpp
```diff
@@ -10,67 +10,79 @@
...
+ * @note Interrupts must be enabled for the UART port in order to use this
+ * function
+ *
+ */
+void CUartCom::startReception()
+{
+    HAL_UART_Receive_IT(mp_huart, &m_rx_char, 1);
+    if (mp_huart->Instance == USART1)
+    {
+        sp_UART_1 = this;
```
Vavat: Ask yourself: does `sp_uart_1` mean that this static pointer belongs to the first instance of `CUartCom` class, or does it mean it belongs to whatever instance is in control of UART1? How does any specific instance of the class know which hardware resource belongs to it? What happens when we want to use UART2, UART3 and UART7 instead?

Vavat: This is possibly my mistake not fully understanding your code. I am not going to delete the comment yet, but don't rush correcting anything.

Vavat: OK. I am pretty sure that what you've done as far as associating numbers of the hardware instances with numbers of the static variables is fine. It's the other stuff I am not so happy about. We can close this comment when others are resolved.

Lib/Src/CUartCom.cpp
```diff
@@ -10,67 +10,79 @@
...
 
 /**
  * @brief Construct UART communication object with half-duplex mode.
- *
  * @param p_huart Pointer to UART hardware control register structure.
+ * @param name Name of the CUartCom Instance
  * @param uart_de_port Pointer to GPIO port.
  * @param uart_de_pin GPIO pin.
  */
 CUartCom::CUartCom(UART_HandleTypeDef *p_huart,
```
Vavat: This is going to fail if this class is instantiated as a global variable on the heap instead of on the stack as part of a controller. It is best to not rely on `p_huart` being defined. 
If you don't understand the differences I am talking about here, ask me and I'll explain.

Lib/Src/CUartCom.cpp
```diff
@@ -98,8 +106,67 @@ void CUartCom::sendMessage(const std::string &msg)
     HAL_UART_Transmit(mp_huart, (uint8_t *)c_msg, msg_len, UART_TIMEOUT);
```
Vavat: This command is blocking. It will take as long to execute as it takes to actually transmit the message. It's a waste of resources. Have a go at using DMA transmission for this purposes. However, using DMA opens a whole bunch of other problems like concurrent access, so passing by reference without copying the messages is not going to work. ;-)

Vavat: Let's discuss if you are not clear about why this is a problem.

Lib/Src/CUartCom.cpp
```diff
@@ -98,8 +106,67 @@ void CUartCom::sendMessage(const std::string &msg)
...
+    return false;
+}
+
+/**
+ * @brief Gets and deletes the first command in the queue
+ * @return Command as string
+ */
+std::string CUartCom::getCommand()
+{
+    std::string command = m_cmd_queue.front();
```
Vavat: Perhaps worth leaving a note here that `std::string` is returned by value, i.e. it is copied, so even when local copy gets out of scope the value is not destroyed.

Lib/Src/CUartCom.cpp
```diff
@@ -10,67 +10,79 @@
...
+void CUartCom::send(const std::string &msg)
 {
     uint16_t msg_len = msg.length();
 
     // Enable UART_DE Pin
-    if (m_uart_de_port != nullptr)
-    {
-        HAL_GPIO_WritePin(m_uart_de_port, m_uart_de_pin, GPIO_PIN_SET);
-    }
+    m_uart_de_pin.set(GPIO_PIN_SET);
```
Vavat: you should use `true` and `false` since this class is meant to decouple from HAL.

Lib/Src/CGpioWrapper.cpp
```diff
@@ -35,13 +72,20 @@ CGpioWrapper::CGpioWrapper(GPIO_TypeDef *p_port, uint16_t pin)
  */
 bool CGpioWrapper::get()
 {
-    if (HAL_GPIO_ReadPin(mp_port, m_pin) == GPIO_PIN_SET)
+    if (mp_port == NULL)
```
Niuslar: change this to nullptr for consistency

Lib/Src/CGpioWrapper.cpp
```diff
@@ -52,13 +96,29 @@ bool CGpioWrapper::get()
  */
 void CGpioWrapper::set(bool value)
 {
-    HAL_GPIO_WritePin(mp_port, m_pin, (value ? GPIO_PIN_SET : GPIO_PIN_RESET));
+    if (mp_port == NULL)
```
Niuslar: Same here and other NULL

Lib/Inc/CUartCom.h
Niuslar: Are static arrays different? I tried using 
``` CUartCom *CUartCom::sp_uart[] = {nullptr}; ```
and it seems to work.

**Comment by Niuslar** [comment]
@Vavat I think I addressed all the points we discussed the other day.

**Comment by Vavat** [comment]
@Niuslar, I asked @Masitech to review as well.

**Review by Vavat — COMMENTED** [review]
Let's merge latest develop into this branch to make sure pipeline works.

File: Lib/Src/CUartCom.cpp [hunk 13]
```diff
@@ -10,96 +10,231 @@
...
+        }
+        else
+        {
+            // Enable USART_DE pin
+            m_uart_de_pin.set(true);
+
+            // Start transmission with interrupts
+            std::string message = m_tx_queue.front();
+            uint8_t msg_length = message.length();
+            strcpy(m_tx_buffer, message.c_str());
```
Vavat: Why do you copy the message out of the buffer? Why not use the buffer as is, and then pop() the queue when transmission is complete?

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
+    HAL_UART_Receive_IT(p_huart, &m_rx_char, 1);
+}
+
+/**
+ * @brief UART Transmission Complete Interrupt Handler
+ * @arg p_huart Pointer to UART Handler
+ */
+void CUartCom::uartTxHandler(UART_HandleTypeDef *p_huart)
+{
+    // Check if queue is empty
```
Vavat, line 239: You can pop() the queue here. For example: `m_tx_buffer` can be turned into a pointer and if it's nullptr, then no need to pop(). If it is not empty, then pop the m_tx_queue and then set m_tx_buffer to nullptr.

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
+    if (m_tx_queue.empty())
```
Vavat: This whole `if()` statement reads really weird. If `m_tx_buffer` is empty, then send empty string. This points towards you loading too much functionality into `send()` method. Why not create a new method called `finishTx()`, move excess functionality from `send()` into new method and call that from here. Then the code will read much nicer.

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
+}
 
-        return;
+/**
+ * @brief Check if the queue has any commands to read
+ * @return True if the queue is not empty
+ */
+bool CUartCom::isCommandAvailable()
+{
+    if (!m_rx_queue.empty())
```
Vavat: alternative syntax. does this read easier?
```
return (m_rx_queue.empty() == false)
```

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
-            // TODO: cannot hardcode stuff like this into classes.
-            //            HAL_GPIO_TogglePin(ALARM_GPIO_Port, ALARM_Pin);
-            HAL_Delay(200);
+            b_success = false;
+            return b_success;
         }
     }
+
+    // Assign hardware to instance
+    sp_UART[s_uart_instances] = this;
```
Vavat: Maybe need to check that you have not exceeded number of instances allowed.

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
+
+/**
+ * @brief Enables UART reception with interrupts
+ * @note UART Interrupts must be enabled in the hardware configuration in order
+ * to use this method
+ *
+ */
+void CUartCom::startRx()
+{
+    HAL_UART_Receive_IT(mp_huart, &m_rx_char, 1);
```
Vavat, line 88: Is it worth making sure that DE line is in the right state? Also, do you switch off Rx IRQ when transmitting?

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
 }
 
 /**
  * @brief sends message via UART.
  * @param msg Message to send.
+ * @return True if there was space in the transmission queue
  */
-void CUartCom::sendMessage(const std::string &msg)
+void CUartCom::send(const std::string msg)
```
Vavat: You need to disable Rx interrupt when transmitting. Otherwise, on shared bus, Receiver might hear the transmission.

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,231 @@
...
+ * to use this method
+ *
+ */
+void CUartCom::startRx()
+{
+    HAL_UART_Receive_IT(mp_huart, &m_rx_char, 1);
 }
 
 /**
  * @brief sends message via UART.
```
Vavat: What exactly does this method do? What happens when the msg is empty? It seems like there is hidden functionality that is not described in this comment.

Lib/Inc/CCmdBuffer.h
```diff
@@ -0,0 +1,34 @@
+/**
+ * @file CCmdBuffer.h
```
Vavat: I don't really understand why this is called CCmdBuffer. There is nothing in the code that makes it command specific.

Lib/Inc/CUartCom.h
```diff
@@ -11,25 +11,59 @@
...
+        TX
+    };
+    enum uart_events
+    {
+        EMPTY_MSG,
+        NON_EMPTY_MSG
+    };
+    bool mb_state = IDLE;
+    CGpioWrapper m_uart_de_pin;
+    CCmdBuffer m_rx_buffer;
```
Vavat: It just occurred to me that you can use `std::queue<char>` instead of CCmdBuffer. ;-)

Lib/Src/CAdcData.cpp
```diff
@@ -14,7 +14,8 @@
 #define ADC_START_BIT        2
 #define ADC_CONVERSION_COEFF ((float)ADC_VDDA / ADC_RES)
 
-CUartCom uart_adc(&huart2);
+// TODO: Delete this instance
+CUartCom uart_adc("ADC Data");
```
Vavat: Should instantiate the `CUartComs` as a global variable in `cpp_main.cpp`. This way they can be linked to `CDispatcher` class instance.

Lib/Src/cpp_link.cpp
```diff
@@ -25,7 +25,7 @@
  * executing relevant classes can call initialisation methods that will
  * configure the hardware.
  */
-CUartCom g_debug_uart(&huart2);
+CUartCom g_debug_uart("Main");
```
Vavat, line 27: ah. Ignore my previous comment about global instantiation. didn't see this.

Vavat (dismissed):
As per our discussion. ping me when you are done.

Lib/Src/CUartCom.cpp
Niuslar: But then I can't store messages in the TX queue if the UART is busy transmitting.

Lib/Src/CUartCom.cpp
Niuslar: What do you mean with "copy out of the buffer"? I'm retrieving the next string in the queue, then making it a c-type string and copying that into the buffer. I could pop() when tx is complete, but I can't see why.

Lib/Src/CUartCom.cpp
Niuslar: ```m_tx_buffer``` a pointer to what?

Lib/Src/CUartCom.cpp
Niuslar: This method enables the interrupts for reception. If the message is empty it will be processed but not pushed to the queue.

Lib/Inc/CUartCom.h
Niuslar: I kept CCmdBuffer because I'm using it to check the end of string when put() is called.

Lib/Src/CUartBuffer.cpp
```diff
@@ -0,0 +1,70 @@
+/**
+ * @file CFifoBuffer.cpp
```
Niuslar: Need to fix this

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,239 @@
...
+void CUartCom::updateTxBuffer()
+{
+    std::string message = m_tx_queue.front();
+    m_tx_msg_length = message.length();
+    strcpy(m_tx_buffer, message.c_str());
+    m_status = TX;
+    m_tx_queue.pop();
+}
+
+void CUartCom::transmit()
```
Niuslar: I know this is a short method, but I'll keep it in case I need to disable RX Interrupts when transmitting.

Lib/Src/CUartCom.cpp
Niuslar: I will change this.

Lib/Inc/CCmdBuffer.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef CCMDBUFFER_H_
+#define CCMDBUFFER_H_
+
+#include <string>
+#include "main.h"
+
+#define BUF_SIZE 60
+
+class CCmdBuffer
```
MAhmedDev: @Niuslar Using C++ templates is quite common for circular buffers.  The buffer becomes data type agnostic.

Lib/Src/CCmdBuffer.cpp
```diff
@@ -0,0 +1,65 @@
...
+CCmdBuffer::~CCmdBuffer()
+{
+    // TODO Auto-generated destructor stub
+}
+
+/**
+ * @brief Adds character to buffer
+ * @return True if the character is \n or \r to mark end of string
+ */
+bool CCmdBuffer::put(const char data)
```
MAhmedDev: This is susceptible to buffer overwrite. By the time head resets the data at that position must already be read by the software. This may not always be the case. Have a look at how circular buffers / FIFO  works. They are everywhere in the real-world packaged under different names. So worth learning about it.

Lib/Src/CUartCom.cpp
Vavat: I don't follow how storing messages in Tx queue is related to Rx interrupt?
When using half-duplex uart both incoming and outgoing messages go through the same set of wires. Some modern half-duplex transceivers prevent echo by blocking Rx when Tx is active, but not all of them do that. This is why Rx interrupt should be disabled when you are transmitting and when you resume receiving, whatever is in the receive uart should be flushed out.

Lib/Src/CUartCom.cpp
Vavat: Short methods are good.

Lib/Src/CUartCom.cpp
Vavat: This relates to my previous comment about how you organised the retrieval of data from `m_tx_queue`. You `strcpy()` the data out. But you could just use the content of the queue directly and then `pop()` the queue when transmission is complete.
Perhaps, another pair programming session is in order. Let me know if you want it.

Niuslar: I can't use the data directly because the queue stores C++ strings and I need to send c-type strings. That's why I'm copying the data to the buffer. Another pair programming session would be great, the last one was very helpful.

Lib/Inc/CCmdBuffer.h
Niuslar: Thanks, I haven't worked much with templates yet. I'll have a look at this.

Lib/Src/CUartCom.cpp
Niuslar: I was testing it by echoing what I was sending to the MCU, and when disabling the RX Interrupts when transmitting, some messages were not being stored in the RX Queue. You're right, maybe there could be a check to disable RX interrupts only when it is half-duplex? Because if the MCU is busy transmitting a bunch of data, then it will not be able to store incoming commands (using half-duplex).

Lib/Inc/CFIFOBuffer.h
```diff
@@ -0,0 +1,53 @@
...
+ *  Created on: Mar 28, 2022
+ *      Author: niuslar
+ */
+
+#ifndef CFIFOBUFFER_H_
+#define CFIFOBUFFER_H_
+
+#include "main.h"
+
+template <typename T, uint8_t BUF_SIZE>
```
MAhmedDev: uint8_t limits buffer size to 255 items. 
Suggestions : unit32_t or uint16_t

Niuslar: Thanks, fixed

Niuslar:
@Vavat, @Masitech, I changed the CUartBuffer class to be a generic FIFO Buffer that accepts any type. I'm using this class for the UART Buffer and the RX and TX string queues. 
I tested the code and it works, but maybe there are some weaknesses that I'm not seeing. I would appreciate it if you could have a look.

Masitech:
> @Vavat, @Masitech, I changed the CUartBuffer class to be a generic FIFO Buffer that accepts any type. I'm using this class for the UART Buffer and the RX and TX string queues. I tested the code and it works, but maybe there are some weaknesses that I'm not seeing. I would appreciate it if you could have a look.

What are your test cases?
Have you tested for some edge cases & extreme value that is where most software fails?

Niuslar:
> What are your test cases? Have you tested for some edge cases & extreme value that is where most software fails?

I tried echoing what I send to the MCU via UART. That way I test RX and TX. The way I'm trying to stress the systems is by sending long pieces of text. Here's the code that I'm using.I know it's a very basic test, any suggestions are welcome. 

``` 
        g_debug_uart.init(&huart2);
        g_debug_uart.startRx();

        // Infinite Loop
        while (1)
        {
            if (g_debug_uart.isCommandAvailable())
            {
                g_debug_uart.send("Com 2:  ");
                g_debug_uart.send(g_debug_uart.getCommand());
                g_debug_uart.send("\n");
            }
        }
```

Niuslar:
I think I addressed all the comments and points discussed during pair programming. It should be safe to merge.

Niuslar:
@Vavat I changed some methods in the CUartCom class and I think it looks much better now. I'm sure there are things that could be improved, but I would like to work on that once this branch is merged, I think it's getting too big and hard to track changes. 
Let me know what you think.

Lib/Src/CUartCom.cpp
```diff
@@ -10,96 +10,339 @@
...
+            mb_half_duplex = true;
+        }
+        m_uart_de_pin.init(uart_de_port, uart_de_pin);
+
+        // Check this hardware is not assigned to another instance
+        int8_t index = CUartCom::getIndex(p_huart);
+        if (index < 0)
+        {
+            // Assign hardware to instance
+            if (s_uart_instances < MAX_UART_ENGINES)
```
Niuslar, line 69: This could probably be just one `if` statement
```if((index < 0) && (s_uart_instances < MAX_UART_ENGINES))```

MAhmedDev (approved):

PR #54, opened 2022-04-12
Add Embedded Template Library (ETL)

### LINKED ISSUE #53: Change std::string to c-style strings
CPP-style strings use dynamic memory allocation which does not follow MISRA standards. 
Need to refactor everything to use C-style strings.

Niuslar:
Thanks @FrozenOxide  for the suggestion. 
The ETL includes strings with static memory allocation. 
I changed all the files that had the ```#include <strings>``` directive to ```#include "../etl/string.h"``` and added 
```#define MAX_STRING_SIZE 60```. 
Now all the strings have a size of 60, but this can be easily changed in each header file that includes strings. 
I applied this changes to my feature/nu_uart_f411 branch and it still works when mirroring messages sent to the MCU, so I guess I didn't break anything too noticeable. 
Let me know what you think. 

Fix #53


MAhmedDev (approved):
Looks good. Thanks for integrating the lib with the codebase.

PR #62, opened 2022-04-20
refactor: rename isCommandAvailble() and getCommand() 

Niuslar:
To make the IComChannel more generic, the methods isCommandAvailable() and getCommand() are now isDataAvailable() and getData()

### DIFF
--- a/Lib/Src/CUartCom.cpp
+++ b/Lib/Src/CUartCom.cpp
@@ -272,7 +272,7 @@ etl::string<MAX_STRING_SIZE> CUartCom::getString()
 }
 
 /**
- * @brief Check if the queue has any commands to read
+ * @brief Check if RX Queue is empty
  * @return True if the queue is not empty
  */
 bool CUartCom::isDataAvailable()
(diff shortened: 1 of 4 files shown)


Lib/Src/CUartCom.cpp
```diff
@@ -275,7 +275,7 @@ etl::string<MAX_STRING_SIZE> CUartCom::getString()
  * @brief Check if the queue has any commands to read
  * @return True if the queue is not empty
  */
-bool CUartCom::isCommandAvailable()
+bool CUartCom::isDataAvailable()
```
Vavat, line 278: Does doxygen header need updating? It still says "command", while the method name is `isDataAvailable()`.
Should probably also define what it means that data is available.

Vavat (commented):
Just a minor comment on the doxygen header for refactored method.

Lib/Src/CUartCom.cpp
Niuslar: Fixed.

Vavat (approved):

PR #65, opened 2022-04-21
feature: add isBusy() method to CFifoBuffer

### LINKED ISSUE #64: CFIFOBuffer race condition issue
If an interrupt calls get() or put() while main() is running one of those methods, the buffer stops working properly.  
Add a m_busy_flag to make it easier to check if the buffer is busy or not.

Niuslar:
Fix #64


Vavat:
Could you describe how you see this functionality being used to solve #64?

Niuslar:
> Could you describe how you see this functionality being used to solve #64?

@Vavat No, I thought I had solved the problem but I realised I didn't. Need to keep working on this.

PR #67, opened 2022-04-22
Feature/nu uart comms

### LINKED ISSUE #63: Refactor CUartCom class to make it more generic
The CUartCom class is currently processing data as strings and storing the strings in queues. 
Some of this functionality could be part of the CParser class and the CUartCom class could focus on just storing the incoming data in a uint8_t buffer. 
I think this could make the code easier to read and more flexible.

Niuslar:
Fix #63 and #64. 
- The CUartCom class now works with bytes instead of strings. 
- Overloaded send() method works with strings or buffer of uint8_t
- UartRxHandler() method is significantly simpler

### DIFF
--- a/Lib/Inc/CFIFOBuffer.h
+++ b/Lib/Inc/CFIFOBuffer.h
@@ -28,6 +28,10 @@ public:
     bool put(T data)
     {
         bool b_success = false;
+        /**
+         * @note Disable interrupts while calling put() to avoid simultaneous
+         * access to the buffer and its variables.
+         */
         __disable_irq();
         if (m_count < BUF_SIZE)
         {
--- a/Lib/Inc/CFIFOBuffer.h
+++ b/Lib/Inc/CFIFOBuffer.h
@@ -10,6 +10,8 @@
 #ifndef CFIFOBUFFER_H_
 #define CFIFOBUFFER_H_
 
+#include "main.h"
+
 template <typename T, uint16_t BUF_SIZE>
 class CFIFOBuffer
 {
@@ -26,6 +28,7 @@ public:
     bool put(T data)
     {
         bool b_success = false;
+        __disable_irq();
         if (m_count < BUF_SIZE)
         {
             m_rx_buffer[m_head] = data;
@@ -36,6 +39,7 @@ public:
             m_count++;
             b_success = true;
         }
+        __enable_irq();
         return b_success;
     }
 
--- a/Lib/Inc/CUartCom.h
+++ b/Lib/Inc/CUartCom.h
@@ -17,12 +17,15 @@
 #include "IComChannel.h"
 #include "usart.h"
 
-#define UART_TIMEOUT      100
-#define MAX_RX_QUEUE_SIZE 40
-#define MAX_TX_QUEUE_SIZE 40
-#define MAX_UART_ENGINES  8
-#define TX_BUF_SIZE       (MAX_STRING_SIZE + 1)
-#define RX_BUF_SIZE       60
+#define UART_TIMEOUT     100
+#define MAX_UART_ENGINES 8
+
+/**
+*@note TX_BUF_SIZE needs to be greater than MAX_STRING_SIZE
+*/
+#define TX_BUF_SIZE      (MAX_STRING_SIZE + 1)
+#define RX_QUEUE_SIZE    100
+#define TX_QUEUE_SIZE    100
 
 class CUartCom : public IComChannel
 {
@@ -35,7 +38,11 @@ public:
               uint16_t uart_de_pin);
 
     void startRx();
+    void stopRx();
+
+    bool send(uint8_t *p_data_buf, uint32_t len);
     bool send(const etl::string<MAX_STRING_SIZE> msg);
+
     bool isDataAvailable();
     uint8_t getData();
 
@@ -43,13 +50,11 @@ public:
     void uartTxHandler(UART_HandleTypeDef *p_huart);
 
     static int8_t getIndex(UART_HandleTypeDef *p_huart);
-
     static CUartCom *sp_UART[MAX_UART_ENGINES];
     static uint8_t s_uart_instances;
     UART_HandleTypeDef *mp_huart;
 
 private:
-    void updateTxBuffer();
     void endTx();
     bool transmit();
     enum uart_status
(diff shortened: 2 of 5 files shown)


Lib/Src/CDispatcher.cpp
```diff
@@ -205,8 +205,9 @@ void CDispatcher::processComChannels()
         while (mp_comchannels[channel]->isDataAvailable())
         {
             bool b_command_recognised;
-            etl::string<MAX_STRING_SIZE> command =
-                mp_comchannels[channel]->getData();
+            // TODO: Change this method to work with the future command parser
```
Niuslar, line 208: I had to change this because getData no longer returns a string. I think this should be done by the parser.

Lib/Inc/CFIFOBuffer.h
```diff
@@ -26,6 +28,7 @@ class CFIFOBuffer
     bool put(T data)
     {
         bool b_success = false;
+        __disable_irq();
```
Vavat, line 35: This definitely needs comments. Explain what is happening and why you are doing that. Otherwise, someone might "refactor it to optimise bandwidth".

Lib/Inc/CUartCom.h
```diff
@@ -17,12 +17,15 @@
...
-#define MAX_UART_ENGINES  8
-#define TX_BUF_SIZE       (MAX_STRING_SIZE + 1)
-#define RX_BUF_SIZE       60
+#define UART_TIMEOUT     100
+#define MAX_UART_ENGINES 8
+
+/**
+*@note TX_BUF_SIZE needs to be greater than MAX_STRING_SIZE
+*/
+#define TX_BUF_SIZE      (MAX_STRING_SIZE + 1)
```
Vavat: I recommend making Tx buffer about 500 to 1000 bytes, Rx buffer can be smaller since we are draining it very fast. Say 100-200.

Sometime in the future, we'll implement high water mark for buffers to see how they are utilised and then we'd be able to optimise the size.

Lib/Src/CUartCom.cpp
```diff
@@ -85,23 +85,39 @@ bool CUartCom::init(UART_HandleTypeDef *p_huart,
...
+    HAL_UART_AbortReceive_IT(mp_huart);
+}
+
+/**
+ * @brief Add string to the TX Queue and start transmission if UART is IDLE
  * @param msg Message to send.
- * @return True if message was added to the queue and transmission started
- * successfully.
+ * @return True if message was added to the queue. If transmission is attempted,
+ * it returns true if it was successful.
```
Vavat: Too much meaning to the return variable. Why wouldn't transmission not be successful?

Lib/Src/CUartCom.cpp
```diff
@@ -113,14 +129,38 @@ bool CUartCom::send(etl::string<MAX_STRING_SIZE> msg)
...
 /**
- * @brief Update TX Buffer with next element on queue
+ * @brief Add data to the TX Queue and start transmission if possible
+ * @param Pointer to data buffer
+ * @param Length of the data buffer (in bytes)
+ * @return True if message was added to the queue and transmission started
+ * successfully.
  */
-void CUartCom::updateTxBuffer()
+bool CUartCom::send(uint8_t *p_data_buf, uint32_t len)
```
Vavat, line 118: Why are there two methods that basically do the same thing?

Vavat:
```
bool CUartCom::send(uint8_t *p_data_buf, uint32_t len)
{
    etl::string<max_string_size> tx_string;
    // copy p_data_buf into tx_string
    bool b_success = send(tx_string);
    return b_success;
}
```

Or do it the other way around where `etl::string<>` interface is calling data pointer interface. That might be more efficient if `etl` library has a method to get c-string pointer to internal string structure.

Vavat: The actual send logic should only be implemented in one place. Otherwise, you are running the danger of having to non-identical methods. One causing bugs and one not. Bad for maintainability.

Lib/Src/CUartCom.cpp
```diff
@@ -133,48 +173,33 @@ bool CUartCom::transmit()
...
-            // UART Parity Error Interrupt
-            __HAL_UART_DISABLE_IT(mp_huart, UART_IT_PE);
-
-            // UART Error Interrupt: (Frame error, noise error, overrun
-            // error)
-            __HAL_UART_DISABLE_IT(mp_huart, UART_IT_ERR);
-
-            // UART Data Register not empty Interrupt
-            __HAL_UART_DISABLE_IT(mp_huart, UART_IT_RXNE);
+            stopRx();
```
Vavat, line 157: This is very nice encapsulation.

Lib/Src/CUartCom.cpp
```diff
@@ -113,14 +129,38 @@ bool CUartCom::send(etl::string<MAX_STRING_SIZE> msg)
...
-void CUartCom::updateTxBuffer()
+bool CUartCom::send(uint8_t *p_data_buf, uint32_t len)
 {
-    etl::string<MAX_STRING_SIZE> message = m_tx_queue.get();
-    m_tx_msg_length = message.length();
-    strcpy(m_tx_buffer, message.c_str());
-    m_status = TX;
+    bool b_success = false;
+
+    // Data object to be added to the queue
```
Vavat: superfluous comments. They create an impression that code is well commented, but don't actually contribute to clarity. Try commenting logical segments. 
Example:
```
// Convert data from pointer format to format required by the hardware interface.
...

// Prepare hardware for transmission
...

// Start transmit if not transmitting already

```
This way comments tell the story of what this code does.

Lib/Src/CUartCom.cpp
```diff
@@ -189,50 +214,18 @@ void CUartCom::endTx()
...
-                m_rx_queue.put(rx_string);
-            }
-        }
-        len_counter = 0;
-    }
-    else
-    {
-        m_rx_buffer.put((char)m_rx_char);
-        len_counter++;
+        send("[ERROR]: Data exceeds max. buffer capacity\n");
```
Vavat: Misleading error message. It's not that data exceeds buffer, it's buffer overflow. Be very precise with your error messages. You might be cursing yourself in one year when you receive this message during debug and start shortening the messages, when in reality the problem is that either the buffer is too small or you're sending data faster than your system can process it.

Vavat: This might sound like nitpicking now, but I promise you'll be very thankful for good error messages when you are on the receiving end of them. When writing those, think about a person who knows nothing about how internal structure of the code works.

Lib/Src/CUartCom.cpp
```diff
@@ -286,16 +255,12 @@ bool CUartCom::isDataAvailable()
 
 /**
  * @brief Get and delete the first element in the RX queue
- * @return Data as string
+ * @return byte of data
  */
-etl::string<MAX_STRING_SIZE> CUartCom::getData()
+uint8_t CUartCom::getData()
```
Vavat, line 239: Why? I fail to see the point of returning a single byte.

Vavat (commented):
A lot of comments. I am not sure what is the purpose of this MR, so slightly confused.

Lib/Src/CUartCom.cpp
Niuslar: You're right,  I forgot to update the error message when I changed the `if` statement.

Lib/Src/CUartCom.cpp
Niuslar: If I have a JSON parser, maybe I can get a stream of data that does not include a '\0' or '\n' character, but it still has the '}' character. 
If we work with strings, that command will never be recognised, but if we let each parser process the data this could work. 
As I said earlier, I don't think UART should process anything, just store all the incoming data.

Niuslar:
I regret this, unless we make substantial changes to the software architecture, I can't see the need to return bytes instead of a string. I will keep some of the changes that I think are needed (like disabling IRQ for the FIFO buffer) and create another MR.

PR #68, opened 2022-04-23
refactor: update the firmware

Vavat:
I want to ensure we all run the same version of the firmware, so this updates to v1.27. As far as I can tell, no actual sourcecode changes. 
It also fixes the name of the .ioc file, which was different from the project name and caused issues when opening it.


Multichannel temperature sensor.ioc
```diff
@@ -363,8 +364,8 @@ ProjectManager.MultiThreaded=true
 ProjectManager.NoMain=false
 ProjectManager.PreviousToolchain=STM32CubeIDE
 ProjectManager.ProjectBuild=false
-ProjectManager.ProjectFileName=Multichannel temperature sensor CPP.ioc
-ProjectManager.ProjectName=Multichannel temperature sensor CPP
+ProjectManager.ProjectFileName=Multichannel temperature sensor.ioc
```
Niuslar, line 367: Why do we need to change this? I have another folder called `Multichannel temperature sensor` from when I was working with C instead of C++. I'd rather keep the name with the CPP in the end, otherwise I can't use this file without moving folders/changing names, etc. I don't mind doing that, but I would like to know why.

Niuslar (commented):
I get an error when I try to access the Device Configuration Tool with the new .ioc file because of the name change. I'd rather keep the old name if possible. 
image

Niuslar:
I managed to make it work. If you're comfortable with this MR feel free to merge.

Niuslar (approved):

Vavat:
@Niuslar, I don't understand why you approved if this breaks your build. Did you fix it on your end?

Niuslar:
> @Niuslar, I don't understand why you approved if this breaks your build. Did you fix it on your end?

Yes. I had to update my IDE and change some folders, but it no longer breaks my build.

PR #70, opened 2022-04-26
fix: typo in filename.

Vavat:
This was annoying me, so I decided to fix it.


Niuslar:
> This was annoying me, so I decided to fix it.

If the library was working fine with that typo then this file is useless, we could delete it.

Niuslar (approved):

Issue #69, opened 2022-04-26
Command parser interface class


Vavat:
Need to rigidly define API for any command parser that we implement regardless of how actual command string arrives. This will allow commands to be passed from dispatcher to controllers.

Niuslar:
I think we can close this

PR #71, opened 2022-04-26
Feature/srm #69 command parser interface class

Vavat:
#69 
Define API to any command object that is to be passed to dispatcher and then to controllers for processing.


Lib/Inc/ICommand.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual unsigned int getArgumentCount() const = 0;
+
+    /**
+     * @brief Access parsed arguments by index of their position in the command
+     * string.
+     *
+     * @param index Position of the argument to be accessed. Zero-based index.
+     * @return Value of the argument at the requested index.
+     */
+    virtual float operator[](unsigned int index) = 0;
```
Niuslar, line 47: I don't really like having three methods to get the command data. Maybe we could use a command_t type and then have just one method getCommand()?
```
    typedef struct command
    {
        etl::string<MAX_STRING_SIZE> name;
        float argument[MAX_ARGUMENTS];
        int argument_count;
    } command_t;
```

Lib/Inc/ICommand.h
```diff
@@ -0,0 +1,69 @@
...
+
+class ICommand
+{
+public:
+    ICommand();
+
+    typedef enum STATUS_T
+    {
+        COMMAND_OK = 0,
+        ERROR_INVALID_NAME,
```
Niuslar: what makes a name invalid? 
I thought this was for the controller to decide, if I the parser gets:
```
>setTem;erature(arg1,arg2);
```
Will this set this status?

Lib/Src/CController.cpp
```diff
@@ -78,10 +79,11 @@ void CController::run()
 /**
  * @brief Give a new command to controller.
  *
- * @param p_command Pointer to NULL terminated command string.
+ * @param p_command Pointer to a command object that contains command name and
+ * arguments.
```
Niuslar, line 83: I don't understand this part. I thought ICommand was an interface for the parser. 
`class CJsonParser : public ICommand`. 
But if it is like that, I can't see the point of passing a pointer to a CJsonParser instance. 
The way I see it, it would be easier to pass a `command_t command` retrieved with `getCommand()` from a CJsonParser instance owned by the Dispatcher. 
Let me know if that makes sense.

Niuslar: Just to expand a bit, in CDispatcher, the newCommand() method would make a call to the parser. If there is more than one, then a call to all the parsers until a COMMAND_OK status is received. 
```
bool CDispatcher::newCommand(etl::string<MAX_STRING_SIZE> string)
{
    bool b_success = false;
    if(json_parser.parse(string) == COMMAND_OK)
    {
        b_success = true;
    }
   return b_success
}
```
Then, back in the processComChannels method, use getCommand and pass it to all the controllers.

Lib/Src/CController.cpp
```diff
@@ -78,10 +79,11 @@ void CController::run()
...
  * @return True if command was recognised.
  */
-bool CController::newCommand(etl::string<MAX_STRING_SIZE> command, IComChannel *p_comchannel)
+bool CController::newCommand(ICommand *p_command, IComChannel *p_comchannel)
```
Niuslar, line 86: `bool CController::newCommand(ICommand::command_t command)`
That's the way I see this working, then the controller just compares `command.name` to a set table of commands.

Niuslar (commented):
I think we have different views on how the command parser should work. I left some comments, let me know if they make sense to you.

Lib/Inc/ICommand.h
```diff
@@ -0,0 +1,69 @@
...
+        ERROR_MALFORMATTED
+    } status_t;
+
+    /**
+     * @brief Get status of command.
+     * @note Command is considered valid if status is returned as COMMAND_OK.
+     *
+     * @return Last parsing status.
+     */
+    virtual status_t getStatus() const = 0;
```
Niuslar: If I understood correctly, this is not necessary.

Niuslar (approved):

PR #73, opened 2022-04-28
refactor: add stopRx() method

Niuslar:
Add stopRx method to make things easier to read. This method uses a HAL function that disables UART RX interrupts.

### DIFF
--- a/Lib/Src/CUartCom.cpp
+++ b/Lib/Src/CUartCom.cpp
@@ -89,6 +89,15 @@ void CUartCom::startRx()
 }
 
 /**
+ * @brief Disable UART RX Interrupts
+ * @note used when using half-duplex
+ */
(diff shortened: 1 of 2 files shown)


Lib/Src/CUartCom.cpp
```diff
@@ -133,17 +142,9 @@ bool CUartCom::transmit()
     if (m_tx_queue.size() > 0)
     {
         // If working with half-duplex, disable RX interrupts
-        if (mb_half_duplex)
+        if (mb_half_duplex && m_status == IDLE)
```
Niuslar, line 145: To avoid calling this function each time transmit is called, I added a condition to check the m_status. If the status is TX, the RX interrupts were already disabled.

Vavat (approved):

PR #75, opened 2022-04-28
Feature/nu json parser

Niuslar:
This is a partial JSON parser. 
It works with one of the following formats: 
{"commandName":[arg1,arg2,arg3,arg4]} for multiple arguments  or
{"commandName:arg1}   when there's only one argument. 

This parser does not support nested objects, that is why I call it a "partial" JSON parser. 

The parser follows this state diagram: 
parser_states(1)

As the note says, if the final state is not COMMAND_OK, it means there was an error and the parser will return the last state it reached. 
I tested it with a list of commands and I think it is easy to identify what is wrong with the command based on the last state. 

Here is the list of commands I tried and the results I got:
## Commands: 
```
{"command1":23}				 
{"command2":44.3}
{"command3":[22,44,77]}
{"command4":[33,44,55,66.5,77.5]}
{"command5":[33 ,44 ,55 ,66.5 ,77.5 ]} 
{command6:45}
{"command7":[55,77,33,55,33,22,34,4,3,23,2]}
{"command8":
>command9(12,53,21);
{"command10":{"command11":[10,34]}}
{"command12":"33,22,1134"}
```

## Results:
```
Command name: command1
Argument 1: 23.0
Command name: command2
Argument 1: 44.3
Command name: command3
Argument 1: 22.0
Argument 2: 44.0
Argument 3: 77.0
Command name: command4
Argument 1: 33.0
Argument 2: 44.0
Argument 3: 55.0
Argument 4: 66.5
Argument 5: 77.5
Command name: command5
Argument 1: 33.0
Argument 2: 44.0
Argument 3: 55.0
Argument 4: 66.5
Argument 5: 77.5
BEGIN_KEY
ERROR_MAX_ARGS
BEGIN_VALUE
IDLE                                                                                                  
ERROR_NESTED_JSON                                                                                     
BEGIN_VALUE 
```

## Note: 
While testing I noticed the following bug:
- If I send the commands one by one, I get the expected results, but if I send all the commands at once, the results get scrambled.
- Example: 
```
Command name: command1
Argument 1: 23.0
Command name: command2
Argument 1: 44.3
Command name: command3
Argument 1: 22.0
Argument 2: 44.0
Argument 3: 77.0
Command name: command4
Argument 1: 33.0
Argument 2: 44.0
Argument 3: 55.0
Argument 4: 66.5
Command name: command5
Argument 1: 33.0                       <-- Command5 should have 5 arguments, not 1
77.5                      <-- This is the 5th argument of command5
BEGIN_KEYERROR_MAX_ARGS                 <-- Missing new-line character
BEGIN_VALUE
IDLE
ERROR_NESTED_JSON
BEGIN_VALUE              
```
I'm including the following code in cpp_link.cpp to get those results:
```
// New includes
#include "CParser.h"
#include "../etl/to_string.h"

//Then in cpp_main() I'm adding:

        g_debug_uart.init(&huart2);
        g_debug_uart.startRx();

        // Infinite Loop
        while (1)
        {
            if (g_debug_uart.isDataAvailable())
            {
                CParser::parser_state_t state =
                    g_parser.parse(g_debug_uart.getData());
                if (state == CParser::COMMAND_OK)
                {
                    g_debug_uart.send("Command name: ");
                    g_debug_uart.send(g_parser.getName());
                    g_debug_uart.send("\n");
                    unsigned int arg_count = g_parser.getArgumentCount();
                    for (unsigned int i = 0; i < arg_count; i++)
                    {
                        g_debug_uart.send("Argument ");
                        etl::string<10> index;
                        etl::to_string((i + 1), index);
                        g_debug_uart.send(index);
                        g_debug_uart.send(": ");
                        etl::string<30> argument;
                        etl::to_string(g_parser[i],
                                       argument,
                                       etl::format_spec().precision(1),
                                       true);
                        g_debug_uart.send(argument);
                        g_debug_uart.send("\n");
                    }
                }
                else
                {
                    g_debug_uart.send(parsing_state[state]);
                    g_debug_uart.send("\n");
                }
            }
        }
```

My guess is that this is a bug in the CUartCom class. Need further testing.

### DIFF
--- a/Lib/Inc/CParser.h
+++ b/Lib/Inc/CParser.h
@@ -50,7 +50,9 @@ public:
         ADD_VALUE,
         VALUE_ADDED,
         END_VALUES,
-        PARSING_OK,
+        COMMAND_OK,
+        ERROR_MAX_ARGS,
+        ERROR_NESTED_JSON,
         ERROR
     } parser_state_t;
 
--- a/Lib/Inc/CParser.h
+++ b/Lib/Inc/CParser.h
@@ -78,12 +78,13 @@ private:
     void getTokens(const etl::string<MAX_STRING_SIZE> &string);
     void endToken(CParser::token_t &token);
 
-    // Private variables
+    // Private member variables
     uint8_t m_argument_counter = 0;
-    etl::vector<token_t, MAX_TOKENS> m_tokens;
-    uint16_t m_token_counter = 0;
     etl::string<MAX_STRING_SIZE> m_command_name;
     float m_arguments[MAX_ARGUMENT_COUNT];
+
+    etl::vector<token_t, MAX_TOKENS> m_tokens;
+    uint16_t m_token_counter = 0;
     parser_state_t m_status = IDLE;
 };
 
--- a/Lib/Inc/CParser.h
+++ b/Lib/Inc/CParser.h
@@ -37,11 +37,10 @@ public:
         START_ARRAY_OP,
         END_ARRAY_OP,
         ARGUMENT_DELIMITER,
-        KEY_VALUE_DIV,
-        INVALID
+        KEY_VALUE_DIV
     } token_type_t;
 
-    typedef enum PARSING_STATUS
+    typedef enum PARSING_STATE
     {
         IDLE,
         BEGIN_KEY,
@@ -53,7 +52,7 @@ public:
         END_VALUES,
         PARSING_OK,
         ERROR
-    } parser_status_t;
+    } parser_state_t;
 
     typedef struct token
     {
@@ -64,7 +63,7 @@ public:
 
     // Public methods
     CParser();
-    CParser::parser_status_t parse(const etl::string<MAX_STRING_SIZE> &string);
+    CParser::parser_state_t parse(const etl::string<MAX_STRING_SIZE> &string);
     etl::string<MAX_STRING_SIZE> getName() const;
     unsigned int getArgumentCount() const;
     float operator[](unsigned int index);
@@ -83,7 +82,7 @@ private:
     uint16_t m_token_counter = 0;
     etl::string<MAX_STRING_SIZE> m_command_name;
     float m_arguments[MAX_ARGUMENT_COUNT];
-    parser_status_t m_status = IDLE;
+    parser_state_t m_status = IDLE;
 };
 
 #endif /* CPARSER_H_ */
--- a/Lib/Inc/CParser.h
+++ b/Lib/Inc/CParser.h
@@ -1,0 +1,89 @@
+/**
+ * @file CParser.h
+ *
+ */
+/*
+ *  Created on: Apr 4, 2022
+ *      Author: niuslar
+ */
+
+#ifndef CPARSER_H_
+#define CPARSER_H_
+
+#include "../etl/string.h"
+#include "../etl/vector.h"
+#include "CFIFOBuffer.h"
+#include "IComChannel.h"
+#include "ICommand.h"
+#include "main.h"
+
+#define MAX_TOKENS 60
+
+#ifndef MAX_STRING_SIZE
+#    define MAX_STRING_SIZE 100
+#endif
+
+class CParser : public ICommand
+{
+public:
+    typedef enum TOKEN_TYPES
+    {
+        WHITESPACE,
+        INTEGER,
+        FLOAT,
+        STRING_LITERAL,
+        BEGIN_OPERATOR,
+        END_OPERATOR,
+        START_ARRAY_OP,
+        END_ARRAY_OP,
+        ARGUMENT_DELIMITER,
+        KEY_VALUE_DIV,
+        INVALID
+    } token_type_t;
+
+    typedef enum PARSING_STATUS
+    {
+        IDLE,
+        BEGIN_KEY,
+        END_KEY,
+        BEGIN_VALUE,
+        BEGIN_ARRAY,
+        ADD_VALUE,
+        VALUE_ADDED,
+        END_VALUES,
+        PARSING_OK,
+        ERROR
+    } parser_status_t;
+
+    typedef struct token
+    {
+        token_type_t type = WHITESPACE;
+        etl::string<MAX_STRING_SIZE> text;
+
+    } token_t;
+
+    // Public methods
+    CParser();
+    CParser::parser_status_t parse(const etl::string<MAX_STRING_SIZE> &string);
+    etl::string<MAX_STRING_SIZE> getName() const;
+    unsigned int getArgumentCount() const;
+    float operator[](unsigned int index);
+
+private:
+    // Private methods
+    void addOperator(CParser::token_t &token,
+                     CParser::token_type_t operator_type,
+                     char character);
+    void getTokens(const etl::string<MAX_STRING_SIZE> &string);
+    void endToken(CParser::token_t &token);
+
+    // Private variables
+    uint8_t m_argument_counter = 0;
+    etl::vector<token_t, MAX_TOKENS> m_tokens;
+    uint16_t m_token_counter = 0;
+    etl::string<MAX_STRING_SIZE> m_command_name;
+    float m_arguments[MAX_ARGUMENT_COUNT];
+    parser_status_t m_status = IDLE;
+};
+
+#endif /* CPARSER_H_ */
--- a/Lib/Inc/ICommand.h
+++ b/Lib/Inc/ICommand.h
@@ -21,7 +21,7 @@
 class ICommand
 {
 public:
-    ICommand();
+    ICommand() {}
 
     /**
      * @brief Get name of the last parsed command.
--- a/Lib/Inc/ICommand.h
+++ b/Lib/Inc/ICommand.h
@@ -28,7 +28,7 @@ public:
      *
      * @return Name of the last parsed command.
      */
-    virtual etl::string<MAX_STRING_SIZE> *getName() const = 0;
+    virtual etl::string<MAX_STRING_SIZE> getName() const = 0;
 
     /**
      * @brief Get number of arguments.
--- a/Lib/Src/CParser.cpp
+++ b/Lib/Src/CParser.cpp
@@ -18,7 +18,7 @@ CParser::CParser() {}
 /**
  * @brief Parse string
  * @param String to be parsed
- * @return Parsing status
+ * @return Last parsing state
  */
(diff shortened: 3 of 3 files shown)


Vavat:
Your conclusion about possible bug in uart is likely correct. I suggest you create a "mirror" and send everything you receive back. then send a bunch of stuff and see if you can detect a pattern of where the error happens.

Niuslar:
The error was that the TX Queue in CUartCom was too small. I changed the size and now it works, altough it breaks again if I send 10x the data all at once.

Vavat:
Could you explain what the symptoms are and a note in the most relevant place explaining why this error happens.

Niuslar:
> Could you explain what the symptoms are and a note in the most relevant place explaining why this error happens.

@Vavat I added an error message. When send() is call too many times, too quickly, there's a buffer overflow. Now when that happens the application sends an error to the uart port saying: "Buffer overflow -> TX Queue". 

I think that should be clear enough in case it happens again? Let me know what you think

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
+/**
+ * @file CParser.h
```
Vavat, line 2: This parser deals with JSON, so it should be reflected in the name. Otherwise, parser that handles plain strings will have to have some odd name.

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+#define CPARSER_H_
+
+#include "../etl/string.h"
+#include "../etl/vector.h"
+#include "CFIFOBuffer.h"
+#include "IComChannel.h"
+#include "ICommand.h"
+#include "main.h"
+
+#define MAX_TOKENS 60
```
Vavat, line 20: What are tokens? This begs for a comment.

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+
+#ifndef MAX_STRING_SIZE
```
Vavat, line 22: Perhaps...?, needs a short comment or a better name.

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+#    define MAX_STRING_SIZE 100
+#endif
+
+class CParser : public ICommand
+{
+public:
+    typedef enum TOKEN_TYPES
```
Vavat, line 29: `TOKEN_TYPES_T`

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+        STRING_LITERAL,
+        BEGIN_OPERATOR,
+        END_OPERATOR,
+        START_ARRAY_OP,
+        END_ARRAY_OP,
+        ARGUMENT_DELIMITER,
+        KEY_VALUE_DIV
+    } token_type_t;
+
+    typedef enum PARSING_STATE
```
Vavat, line 43: `PARSER_STATE_T`

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+    {
+        IDLE,
```
Vavat, line 45: It's a common convention to have IDLE be zero, so 
```
{
    IDLE = 0,
    ...
}
```

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+        ADD_VALUE,
+        VALUE_ADDED,
+        END_VALUES,
+        COMMAND_OK,
+        ERROR_MAX_ARGS,
+        ERROR_NESTED_JSON,
+        ERROR
+    } parser_state_t;
+
+    typedef struct token
```
Vavat, line 59: Structs should follow the same naming conventions as classes, because in essence they are classes. Example, you can have a constructor for a struct.
```
typedef struct SToken
{
    ...
} token_t
```

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+    {
+        token_type_t type = WHITESPACE;
+        etl::string<MAX_STRING_SIZE> text;
+
+    } token_t;
+
+    // Public methods
```
Vavat, line 66: Superfluos.

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+    CParser();
```
Vavat, line 67: ctor should be at the top of the file straight after class definition. This is the thing that is looked up the most often by the user of the class and scrolling down takes time.

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+    CParser::parser_state_t parse(const etl::string<MAX_STRING_SIZE> &string);
+    etl::string<MAX_STRING_SIZE> getName() const;
+    unsigned int getArgumentCount() const;
+    float operator[](unsigned int index);
+
+private:
+    // Private methods
```
Vavat, line 74: It literally says `private:` one line above. ;-) 
These types of comments make code look amateurish.

Lib/Inc/CParser.h
```diff
@@ -0,0 +1,91 @@
...
+    void addOperator(CParser::token_t &token,
+                     CParser::token_type_t operator_type,
+                     char character);
+    void getTokens(const etl::string<MAX_STRING_SIZE> &string);
+    void endToken(CParser::token_t &token);
+
+    // Private member variables
```
Vavat, line 81: Superfluous.

Lib/Inc/ICommand.h
```diff
@@ -21,14 +21,14 @@
...
-    ICommand();
+    ICommand() {}
 
     /**
      * @brief Get name of the last parsed command.
      *
      * @return Name of the last parsed command.
      */
-    virtual etl::string<MAX_STRING_SIZE> *getName() const = 0;
+    virtual etl::string<MAX_STRING_SIZE> getName() const = 0;
```
Vavat, line 31: This will break a lot of code. Why do you want to return by value?

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+ * @file CParser.cpp
+ */
+/*
+ * Created on : Apr 4, 2022
+ * Author : niuslar
+ */
+
+#include "CParser.h"
+
+#define ERROR_VALUE (0.00)
```
Vavat, line 11: Bad name. That's not an error value, but more like a `DEFAULT_VALUE`. No?

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+/**
+ * @brief Constructor
+ */
+CParser::CParser() {}
+
+/**
+ * @brief Parse string
+ * @param String to be parsed
+ * @return Last parsing state
```
Vavat, line 21: This is very vague and requires more explanation. Why are you returning last parsing state? How is the user meant to utilise it?

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+    m_command_name = "";
+
+    // Get tokens from string
+    getTokens(string);
+
+    for (uint16_t i = 0; i < m_token_counter; i++)
+    {
+        token_t current_token = m_tokens[i];
+
+        switch (current_token.type)
```
Vavat, line 38: Have you considered building state machine based on parser state, not current token? The way you did it is rather unconventional.

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+                break;
+            case KEY_VALUE_DIV:
+                if (m_status == END_KEY)
+                {
+                    m_status = BEGIN_VALUE;
+                }
+                break;
+            case INTEGER:
+            case FLOAT:
+                if (m_status == BEGIN_VALUE)
```
Vavat, line 78: All of this inside `case FLOAT:` must go into a separate method and get called.

Lib/Src/CParser.cpp
Vavat: This state machine is very hard to follow. The state changes depending on the value of the token, but your implemented it as if token change depending on state.

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+            default:
+                m_status = ERROR;
+        }
+    }
+
+    return m_status;
+}
+
+/**
+ * @brief Process string to get tokens
```
Vavat, line 121: What does it mean `tokenised`? This header needs a lot more explanation. You should assume reader does not know how JSON works.

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+ */
+void CParser::getTokens(const etl::string<MAX_STRING_SIZE> &string)
+{
+    // Reset variables
+    m_tokens.clear();
+    m_token_counter = 0;
+    token_t current_token;
+
+    // Loop through characters in string
+    for (char current_char : string)
```
Vavat, line 132: For someone unfamiliar with how range operations work, it's convenient to leave references.
https://www.techiedelight.com/iterate-over-characters-string-cpp/

Lib/Src/CParser.cpp
```diff
@@ -0,0 +1,309 @@
...
+    {
+        switch (current_char)
```
Vavat, line 134: Again, this state machine is inside out. Thus hard to follow. In my opinion. Have you tried drawing it? If you did, I think you'll find it looks much simpler if you process chars based on current state, not process states, based on current char.
For example, this massive column of case statements can turn into `isnum()` call if you process states.

Vavat (commented):
Happy to discuss, but in current state this class is too hard to comprehend.

Vavat:
Implementation of the state machine does not match the state diagram.

Lib/Src/CParser.cpp
Vavat: Much more comprehensive, but really dense reference: https://en.cppreference.com/w/cpp/language/range-for

Lib/Inc/ICommand.h
Niuslar: The way this method is right now doesn't work if I store the name in a `etl::string<MAX_STRING_SIZE> m_command_name;` variable. 
The ways I could make it work were by either removing the pointer or by changing it to: 
`virtual const etl::string<MAX_STRING_SIZE> getName() const = 0` . Perhaps this is a better option.

PR #77, opened 2022-05-02
Feature/srm hardware abstraction layer

Vavat:
This branch creates an abstraction layer separation between core functionality of the code and access to what's outside the MCU. Specifically, two classes with identical API are created that present peripherals interface. Once accesses real hardware, the other mocks the hardware with a first order approximation of ordinary differential equation time domain solution. Sort of a very crude Newton-Raphson method.


Hardware/CMockHardwareMap.h
```diff
@@ -0,0 +1,58 @@
...
+class CMockHardwareMap : public IHardwareMap, public CController
+{
+public:
+    CMockHardwareMap(etl::string<MAX_STRING_SIZE> name, uint32_t run_period_ms);
+    virtual void init();
+    virtual float getInputVoltage() const;
+    virtual float getTotalCurrent() const;
+    virtual float getControlCurrent() const;
+    virtual float getAmbientTemp() const;
+    virtual float getChanneTemp(uint8_t channel) const;
```
Niuslar: I guess this is getChannelTemp instead of getChanneTemp. This typo is in every file.

Hardware/CMockHardwareMap.cpp
```diff
@@ -0,0 +1,297 @@
...
+ */
+bool CMockHardwareMap::newCommand(ICommand *p_command,
+                                  IComChannel *p_comchannel)
+{
+    bool b_command_recognised = false;
+    /**
+     * Command to modify mock hardware ambient temperature
+     * >setambient(20);
+     */
+    if (p_command->getName()->compare("setambient"))
```
Niuslar: I think the compare() method returns 0 if it is true. In that case this should be 
`if (!(p_command->getName()->compare("setambient")))`

Hardware/CMockHardwareMap.cpp
```diff
@@ -0,0 +1,297 @@
...
+    }
+    /**
+     * Command to modify power rating of the heater. Rating refers to heater
+     * full power output. Channel is optional and if omitted command is applied
+     * to all channels.
+     * >setRating(rating, [channel])
+     */
+    if (p_command->getName()->compare("setrating"))
+    {
+        uint8_t channel = (uint8_t)(*p_command)[1];
```
Niuslar, line 212: Since ICommand has a getArgumentCount() method, maybe it is worth adding a check before assigning values. If the argument count doesn't match the number of arguments the method takes then there's something wrong with the command.

Hardware/CMockHardwareMap.cpp
```diff
@@ -0,0 +1,297 @@
...
+    {
+        m_heater_rating[channel - 1] = rating;
+    }
+}
+
+void CMockHardwareMap::setcapacity(float heater_capacity,
+                                   float radiator_capacity,
+                                   uint8_t channel)
+{
+    if (channel == 0)
```
Niuslar, line 358: Below you use `channel - 1`,  I guess it should be the same here: 
`if ( (channel - 1) == 0)`
Unless you are assuming that the parser will return 0 when trying to get an argument that doesn't exist (which I don't think is a good idea).

Hardware/CMockHardwareMap.cpp
```diff
@@ -0,0 +1,297 @@
...
+        }
+        else
+        {
+            setrating(rating, channel);
+        }
+        b_command_recognised = true;
+    }
+    /**
+     * Command to modify heat capacity of the heater. Channel is optional and if
+     * omitted command is applied to all channels.
```
Niuslar, line 235: This is why I think an argument count check could be useful. 
If someone sends a command without the channel argument, this method will try to retrieve a value that doesn't exist (`(uint8_t)(*p_command)[2]`). 

The way my parser handles this is by sending an ERROR_VALUE of 999.9, which is easy to recognise. 
In this case it will only send the `"Arguments out of bounds"` error and stop. 

I also think there should be a note clarifying if channels start from 0 or 1.

Hardware/CMockHardwareMap.cpp
Vavat: Fixed.

Hardware/CMockHardwareMap.cpp
Vavat: Good point. I am still asigning values before testing. Perhaps, have a look with fresh eyes how to sequence multiple if/else statements, yet not assign before check about argument count is made.
All options I tried look really ugly.

Hardware/CMockHardwareMap.cpp
Vavat: There is a convention that non-existent values should be assigned zero. 999.9 is not a good choice because in some circumstances this might be a valid value. 
However, I take your point about checking argument count.

Hardware/CMockHardwareMap.cpp
Vavat: zero means that optional parameter was omitted from the command. If optional parameter is omitted, then command is applied to all channels. This is correct logic.

Hardware/CMockHardwareMap.h
Vavat: Fixed.

Hardware/CMockHardwareMap.cpp
Niuslar: Ok, I didn't know that, I'll fix that in the parser. That's actually the reason I set it to 999.9, because I thought 0 could be an actual value.

Hardware/CMockHardwareMap.cpp
```diff
@@ -0,0 +1,350 @@
...
+                               m_run_period_ms / 1000 / m_incubator_capacity;
+}
+
+/**
+ * @note: This method is not for sending commands like >setHeaterPower(). This
+ * method is to simulate input from real world conditions changing as the device
+ * is being tested. For example: induce failure of a heater or thermistor. Or
+ * set different ambient temperature.
+ */
+bool CMockHardwareMap::newCommand(ICommand *p_command,
```
MAhmedDev, line 179: Using C++ references will make the code more readable. 
If possible we should consider using references over pointers for safety and code readability reasons.

Lib/Src/CAdcData.cpp
```diff
@@ -48,7 +48,7 @@ void CAdcData::init()
  * @param adc_channel Number of channel starting from 0
  * @retval Volts
  */
-float CAdcData::operator[](uint8_t adc_channel)
+float CAdcData::operator[](uint8_t adc_channel) const
 {
     // Check channel is within defined range
```
MAhmedDev, line 53: For line 60 which is not part of this MR however useful to add: 
C++ style casts are checked by the compiler. C style casts aren't and can fail at runtime.

Hardware/CMockHardwareMap.cpp
Niuslar: I think you're right, but this should be applied everywhere. I'll create a ticket to change pointers for references where possible.

Niuslar (approved):

PR #78, opened 2022-05-02
Add jobs to the workflow to download code to the MCU

Niuslar:
I changed the workflow a bit, now it does the following: 
- "build" uploads 2 artifacts: Release build and Debug build (both .bin files)
- If "build" was successful, the next two jobs load the .bin files into the MCU using the Pi. 

The only change I made to the project was that I ticked the option to generate binaries, I don't know why this changed so many lines of code.


Vavat:
@Masitech, you are a much better person to review this. I have no idea what's going on.

Vavat:
Do we need to create a new build profile for HIL?

.github/workflows/c-cpp.yml
```diff
@@ -31,11 +31,35 @@ jobs:
...
         with:
-          name: Build Artifact
-          path: |
-            /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Debug/Multichannel\ temperature\ sensor.bin
-            /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Debug/Multichannel\ temperature\ sensor.hex
-            /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Debug/Multichannel\ temperature\ sensor.elf
-            /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Release/Multichannel\ temperature\ sensor.bin
-            /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Release/Multichannel\ temperature\ sensor.hex
-            /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Release/Multichannel\ temperature\ sensor.elf
+          name: debug_build
```
Masitech, line 34: upload_debug_build
Might be a better name for this stage since it is not building any software.

.github/workflows/c-cpp.yml
```diff
@@ -31,11 +31,35 @@ jobs:
...
+          path: /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Debug/Multichannel\ temperature\ sensor.bin
+      - uses: actions/upload-artifact@v3
+        with:
+          name: release_build
```
Masitech, line 38: upload_release_build

MAhmedDev:
Looks Good, 
Nice Work! 

A minor improvement can be made by removing absolute naming in paths since it will break by using CI variables.

MAhmedDev:
@Niuslar  You are flashing both debug and release versions of the software is that required?

Niuslar:
> @Niuslar You are flashing both debug and release versions of the software is that required?

The idea I had in mind was to run tests for each build, there might be things that work in Debug but not in Release.

MAhmedDev (approved):

Niuslar:
Let's merge this once a permanent Raspberry Pi + Nucleo-64 is setup.

Niuslar:
> A minor improvement can be made by removing absolute naming in paths since it will break by using CI variables.

@Masitech Fixed

PR #79, opened 2022-05-02
Fix/nu uart bufer overflow

Niuslar:
When the send() method is call too many times, too quickly, it can lead to a buffer overflow. 
I increased the buffers  and queues size to prevent this from happening, but it can still happen. 
Because of this I added some error messages to make it easy to identify what's causing problems.

### DIFF
--- a/Lib/Inc/CUartCom.h
+++ b/Lib/Inc/CUartCom.h
@@ -18,10 +18,10 @@
 #include "usart.h"
 
 #define UART_TIMEOUT      100
-#define MAX_RX_QUEUE_SIZE 40
-#define MAX_TX_QUEUE_SIZE 40
+#define MAX_RX_QUEUE_SIZE 50
+#define MAX_TX_QUEUE_SIZE 100
 #define MAX_UART_ENGINES  8
-#define TX_BUF_SIZE       (MAX_STRING_SIZE + 1)
+#define TX_BUF_SIZE       70
 #define RX_BUF_SIZE       60
 
 class CUartCom : public IComChannel
--- a/Lib/Inc/CUartCom.h
+++ b/Lib/Inc/CUartCom.h
@@ -19,7 +19,7 @@
 
 #define UART_TIMEOUT      100
 #define MAX_RX_QUEUE_SIZE 50
-#define MAX_TX_QUEUE_SIZE 100
+#define MAX_TX_QUEUE_SIZE 3000  // size in bytes
 #define MAX_UART_ENGINES  8
 #define TX_BUF_SIZE       70
 #define RX_BUF_SIZE       60
@@ -36,6 +36,7 @@ public:
 
     void startRx();
     bool send(const etl::string<MAX_STRING_SIZE> msg);
+    bool send(uint8_t *p_data_buf, uint32_t len);
     bool isDataAvailable();
     etl::string<MAX_STRING_SIZE> getData();
     void uartRxHandler(UART_HandleTypeDef *p_huart);
@@ -63,11 +64,11 @@ private:
     uint8_t m_status = IDLE;
     CGpioWrapper m_uart_de_pin;
     CFIFOBuffer<char, RX_BUF_SIZE> m_rx_buffer;
-    char m_tx_buffer[TX_BUF_SIZE] = {0};
+    uint8_t m_tx_char;
     uint8_t m_tx_msg_length = 0;
     uint8_t m_rx_char;
     CFIFOBuffer<etl::string<MAX_STRING_SIZE>, MAX_RX_QUEUE_SIZE> m_rx_queue;
-    CFIFOBuffer<etl::string<MAX_STRING_SIZE>, MAX_TX_QUEUE_SIZE> m_tx_queue;
+    CFIFOBuffer<uint8_t, MAX_TX_QUEUE_SIZE> m_tx_queue;
 };
 
 #endif /* CUARTCOM_H_ */
--- a/Lib/Src/CUartCom.cpp
+++ b/Lib/Src/CUartCom.cpp
@@ -109,11 +109,14 @@ bool CUartCom::send(etl::string<MAX_STRING_SIZE> msg)
     // Add message to queue
     if (m_tx_queue.size() <= MAX_TX_QUEUE_SIZE && (msg.empty() == false))
     {
-        if (m_tx_queue.put(msg) == false)
+        if (m_tx_queue.put(msg) == true)
+        {
+            b_success = true;
+        }
+        else
         {
             send("Error: Buffer overflow -> TX Queue\n");
         }
-        b_success = true;
     }
 
     // Start transmission only if UART is idle
--- a/Lib/Src/CUartCom.cpp
+++ b/Lib/Src/CUartCom.cpp
@@ -10,8 +10,6 @@
 
 #include "CUartCom.h"
 
-#define BYTE 1
-
 CUartCom *CUartCom::sp_UART[MAX_UART_ENGINES] = {nullptr};
 uint8_t CUartCom::s_uart_instances = 0;
 
@@ -130,7 +128,6 @@ bool CUartCom::send(uint8_t *p_data_buf, uint32_t len)
         }
     }
 
-    // Start transmission only if UART is not transmitting already
     if (m_status == IDLE)
     {
         transmit();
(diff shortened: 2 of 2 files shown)


Lib/Src/CUartCom.cpp
```diff
@@ -109,7 +109,10 @@ bool CUartCom::send(etl::string<MAX_STRING_SIZE> msg)
     // Add message to queue
     if (m_tx_queue.size() <= MAX_TX_QUEUE_SIZE && (msg.empty() == false))
     {
-        m_tx_queue.put(msg);
+        if (m_tx_queue.put(msg) == false)
+        {
+            send("Error: Buffer overflow -> TX Queue\n");
```
Vavat: In this `send()` method you are finding that Tx queue is full and then call `send()` method to put more stuff in Tx queue.

Vavat: Why not make buffers so big you cannot saturate them sending reasonable amount of data?

MAhmedDev: Suggestion, have a coms status flag (enums)
COMS_OK,
COMS_BUSY,
COMS_OVERFLOW etc 
When overflow happens set the status to COMS_OVERFLOW either wait for it to clear or disallow any new messages to be sent. 

In the past, I used SDO to push error messages and kept the serial UART clean.

Niuslar:
I changed the logic behind send() and transmit() to store data in a big buffer and send one byte at a time.
This works very well and it is simple. 
I also changed the way the uartRxHandler works. The logic before was unnecessarily complicated and had some bugs. 
Regarding the error log in the send() method, I removed it because it didn't make sense to try call send when the TX buffer was full. The return value of that function is false only if there was a buffer overflow, the user can decide what to do with this info.

Lib/Inc/CUartCom.h
```diff
@@ -63,11 +64,11 @@ class CUartCom : public IComChannel
     uint8_t m_status = IDLE;
     CGpioWrapper m_uart_de_pin;
     CFIFOBuffer<char, RX_BUF_SIZE> m_rx_buffer;
-    char m_tx_buffer[TX_BUF_SIZE] = {0};
+    uint8_t m_tx_char;
     uint8_t m_tx_msg_length = 0;
     uint8_t m_rx_char;
     CFIFOBuffer<etl::string<MAX_STRING_SIZE>, MAX_RX_QUEUE_SIZE> m_rx_queue;
-    CFIFOBuffer<etl::string<MAX_STRING_SIZE>, MAX_TX_QUEUE_SIZE> m_tx_queue;
+    CFIFOBuffer<uint8_t, MAX_TX_QUEUE_SIZE> m_tx_queue;
```
Vavat, line 71: `MAX_TX_QUEUE_SIZE` name implies there is `MIN_TX_QUEUE_SIZE`. Perhaps a different name such as `TX_BUFFER_SIZE` would reflect the purpose of this variable better.

Lib/Src/CUartCom.cpp
```diff
@@ -10,6 +10,8 @@
 
 #include "CUartCom.h"
 
+#define BYTE 1
```
Vavat: I realise you are trying to avoid "magic numbers" problem, but in this case, it does not add to clarity, but obfuscates that you are sending one byte at a time.

Lib/Src/CUartCom.cpp
```diff
@@ -98,25 +100,40 @@ void CUartCom::stopRx()
...
+ * @brief send string
+ * @param msg Message to be sent via UART
+ * @return success if message was successfully stored in TX Queue.
  */
 bool CUartCom::send(etl::string<MAX_STRING_SIZE> msg)
 {
-    bool b_success = false;
-    // Add message to queue
-    if (m_tx_queue.size() <= MAX_TX_QUEUE_SIZE && (msg.empty() == false))
+    uint32_t msg_len = msg.length();
```
Vavat, line 107: This is very nice wrapper. Perhaps worth adding an `@note:` to the doxygen header to explain that this is a wrapper for pointer based `send()`.

Lib/Src/CUartCom.cpp
```diff
@@ -98,25 +100,40 @@ void CUartCom::stopRx()
...
-        m_tx_queue.put(msg);
-        b_success = true;
+        if (m_tx_queue.put(p_data_buf[i]) == false)
+        {
+            b_success = false;
+        }
     }
 
-    // Start transmission only if UART is idle
+    // Start transmission only if UART is not transmitting already
```
Vavat: This is really nit picking, but the comment does not match the code. Your code says:
**If idle then transmit**, while your comment says **If not transmitting then transmit**. It is not clear to the code reader that these states are a complete set of possible states. Your code is correct in state transitions. You should go into transmit state only from Idle state, so changing the comment is the best approach.

Vavat: Or I'd remove the comment altogether. It does not add to comprehension. Your code is very well written and does not require any aid to understand.

Lib/Src/CUartCom.cpp
```diff
@@ -98,25 +100,40 @@ void CUartCom::stopRx()
...
     {
-        m_tx_queue.put(msg);
-        b_success = true;
+        if (m_tx_queue.put(p_data_buf[i]) == false)
+        {
+            b_success = false;
+        }
     }
 
-    // Start transmission only if UART is idle
```
Vavat, line 116: haha. just realised you had a better comment before, but changed it for some reason.

Lib/Src/CUartCom.cpp
```diff
@@ -151,7 +166,7 @@ bool CUartCom::transmit()
         m_uart_de_pin.set(true);
 
         updateTxBuffer();
-        HAL_UART_Transmit_IT(mp_huart, (uint8_t *)m_tx_buffer, m_tx_msg_length);
+        HAL_UART_Transmit_IT(mp_huart, &m_tx_char, BYTE);
```
Vavat: Much clearer is you explicitly state that you are sending 1 of `m_tx_char`. The rules about "magic numbers" only applies when it is not clear where the number is coming from and giving it a name is meant to fix the ambiguity. It does not apply here.
```
HAL_UART_Transmit_IT(mp_huart, &m_tx_char, 1);
```

Lib/Src/CUartCom.cpp
```diff
@@ -181,42 +196,25 @@ void CUartCom::endTx()
  */
 void CUartCom::uartRxHandler(UART_HandleTypeDef *p_huart)
 {
-    // Store incoming char
-    static uint8_t len_counter = 0;
+    // Store char in buffer
```
Vavat: You still post superfluous comments that do not actually help with understanding the code.

Lib/Src/CUartCom.cpp
```diff
@@ -181,42 +196,25 @@ void CUartCom::endTx()
...
+    }
 
-    bool full_buffer = (len_counter >= (RX_BUF_SIZE));
     if (m_rx_char == '\n' || m_rx_char == '\r')
     {
-        // '\n' and '\r' are replaced with '\0' to mark the end of the string
-        m_rx_buffer.put('\0');
-
-        if (full_buffer)
+        // Get string and store it in RX QUEUE
```
Vavat: and another superfluos comment.

Lib/Src/CUartCom.cpp
```diff
@@ -181,42 +196,25 @@ void CUartCom::endTx()
...
-            m_rx_buffer.reset();
-        }
-        else
-        {
-            etl::string<MAX_STRING_SIZE> rx_string = getString();
-            if (m_rx_queue.size() <= MAX_RX_QUEUE_SIZE && !rx_string.empty())
+            if (m_rx_queue.put(rx_string) == false)
             {
-                m_rx_queue.put(rx_string);
+                send("Error: Buffer overflow -> RX Queue\n");
```
Vavat, line 215: I would advocate we don't send this type of error back through the uart. I suggest we log it and then dump the log of errors when requested. Talk to @Masitech about how he organised logging in the past.

Lib/Src/CUartCom.cpp
```diff
@@ -247,18 +245,22 @@ etl::string<MAX_STRING_SIZE> CUartCom::getString()
 {
     uint8_t counter = 0;
     char c_string[RX_BUF_SIZE];
-    char data;
+    char data = m_rx_buffer.get();
     /**
      * @note while loop stops when the counter is (RX_BUF_SIZE - 1) because
```
Vavat, line 243: Now this is a very useful comment.

Vavat (approved):
This is very good. I like it. All of the comments are of the kind where I would enforce them in a commercial environment, but in this case I'll leave it up to you whether you want to fix them.

PR #80, opened 2022-05-06
feature: process commands with table

Niuslar:
@Vavat I saw this in a book and thought it could be used here. It is a way to process commands using an array of commands and function pointers. What I did was: 
- Create a command_t struct that includes name, argument count and function to execute
- Create an array of commands 
- Create function wrappers so all of them have the same signature 
- In newCommand I moved the "Argument out of bounds" check inside each method it calls. 
- I created two new member functions for setambient and setincubator, before they were part of the newCommand method

I think this is much easier to read and it makes it easier to add new commands if we need to. 
Let me know what you think


Niuslar:
This could lead to bugs and it is not clear for someone who's not familiar with the code.

PR #83, opened 2022-05-09
Feature/srm temperature controller

Vavat:
This controller is bare bones of what we will need. It does not have a lot of functionality required, but it should work. Merging this will allow us to test hardware maps, dispatcher logic, controller command execution and general architecture.


Controllers/CTemperatureController.cpp
```diff
@@ -0,0 +1,299 @@
...
+     * then apply to all channels. if power argument is negative value,
+     * disable heater override.
+     * @note This command suspends operation of the control on
+     * the channel, but does not disable it. Once override is removed,
+     * channel returns to pre-override state. If it was idle, it returns to
+     * idle, if it was under control, then control execution resumes.
+     *
+     * heater(power, [channel])
+     */
+    if (p_command->getName()->compare("heater"))
```
Niuslar: I think you forgot to add the == 0 after compare(). Compare() returns 0 if true.

Controllers/CTemperatureController.cpp
```diff
@@ -0,0 +1,299 @@
...
+    {
+        result = overrideHeater(p_command);
+        b_command_recognised = true;
+    }
+    if (b_command_recognised)
+    {
+        switch (result)
```
Niuslar, line 112: Maybe in the future newCommand could return this result and the Dispatcher could be in charge of sending these messages. I think it could make it easier if we're adding more controllers.

Lib/Src/cpp_link.cpp
```diff
@@ -33,16 +34,30 @@ CDispatcher g_dispatcher(&g_debug_uart);
...
+ * time between simultaneous requests will be 97 * 89 = 8633ms. This is rare
+ * enough that any contention resulting in delay of execution will not stack up.
+ * For three controllers to have contention over execution time period is even
+ * larger. Example: for 89ms, 97ms, and 101ms to cause contention 871'933ms must
+ * pass, which is equivalent to 14.5 hours. This will be very rare event that is
+ * unlikely to cause any issues with clock starvation. For choices of prime
+ * numbers to use see prime_numbers.txt file.
+ */
+CDebugController g_debug_controller("debug", 97);
+CTemperatureController g_temp_controller("temperature", 89);
```
Niuslar: Very interesting.

Niuslar (commented):
Thanks, very easy to understand. There's only one thing that I think needs to be changed.

Controllers/CTemperatureController.h
```diff
@@ -0,0 +1,41 @@
...
+#define CHANNEL_NUMBER 10
+
+class CTemperatureController : public CController
+{
+public:
+    CTemperatureController(etl::string<MAX_STRING_SIZE> name,
+                           uint32_t run_period_ms);
+
+    virtual void run();
+    virtual bool newCommand(ICommand *p_command, IComChannel *p_comchannel);
```
MAhmedDev, line 30: Using C++ references makes the code more readable & safer.

Hardware/CMockHardwareMap.cpp
```diff
@@ -238,16 +238,16 @@ bool CMockHardwareMap::newCommand(ICommand *p_command,
     {
         switch (result)
         {
-            case COMMAND_OK:
+            case ICommand::COMMAND_OK:
                 p_comchannel->send("OK.\n");
                 break;
-            case ERROR_ARG_COUNT:
+            case ICommand::ERROR_ARG_COUNT:
```
MAhmedDev, line 244: Wouldn't it be cleaner to send a string version of the error enum "ERROR_ARG_COUNT" or "ERROR_OUT_OF_BOUNDS" etc to keep up with consistency in other places where error messages may also be sent?

Hardware/CMockHardwareMap.cpp
```diff
@@ -335,25 +333,25 @@ CMockHardwareMap::command_error_code_t CMockHardwareMap::setRating(
...
     ICommand *p_command)
 {
     // Sanitise command arguments
     if ((p_command->getArgumentCount() < 2) ||
         (p_command->getArgumentCount() > 3))
     {
-        return ERROR_ARG_COUNT;
+        return ICommand::ERROR_ARG_COUNT;
     }
     float heater_capacity = (*p_command)[0];
```
MAhmedDev, line 348: Magic numbers make code harder to read and understand. Arguments can be made to use enums for better readability and safety.

Lib/Src/cpp_link.cpp
MAhmedDev: This is pretty smart!

Controllers/CTemperatureController.cpp
```diff
@@ -0,0 +1,299 @@
...
+ *  Created on: 26 Apr 2022
+ *      Author: salavat.magazov
+ */
+
+#include "CTemperatureController.h"
+#include <stdio.h>
+#include "ICommand.h"
+#include "IHardwareMap.h"
+
+extern IHardwareMap *gp_hardware_map;
```
MAhmedDev: Why not pass this into the class as an argument rather than relying on the linker to figure this out during compile time?

Vavat: Done

Controllers/CTemperatureController.cpp
Vavat: Some controllers might want to return extended messages. It's better to leave this with controllers for now.

Controllers/CTemperatureController.h
Vavat: I don't follow. You suggest this should be:
```
virtual bool newCommand(ICommand &command, IComChannel &comchannel);
```

Hardware/CMockHardwareMap.cpp
Vavat: Don't follow the comment.

Hardware/CMockHardwareMap.cpp
Vavat: Make a suggestion how to use enums.

Controllers/CTemperatureController.cpp
Vavat: Fixed

Hardware/CMockHardwareMap.cpp
MAhmedDev: Since this enum is part of ICommand interface class, other classes will implement this interface to process commands with their own custom message.  Over time this will become difficult to manage since each implementation will have its own unique & hardcoded message also since the message holds the same amount of information as the enum name it makes sense to send a string representation of the ENUM name. This is however not supported by default in C & C++ (unlike python) nevertheless, there are solutions which I think are neat. 

Proposed Solution C & C++
[Example C & C++](https://godbolt.org/z/44hqEef6o )


```
#include <iostream>
#include "magic_enum.hpp" // Header only library

using namespace std;

/*************** C Implementation - Manual Management *************/
enum ErrorType { COMMAND_OK = 0, ERROR_OUT_OF_BOUNDS, ERROR_TYPE_MISMATCH, MAX_ERROR_TYPE}; 

static const char * ErrorTypeName [MAX_ERROR_TYPE] = {
    /* ENUM -> STRING */
    [COMMAND_OK] = "COMMAND_OK",
    [ERROR_OUT_OF_BOUNDS] = "ERROR_OUT_OF_BOUNDS",
    [ERROR_TYPE_MISMATCH] = "ERROR_TYPE_MISMATCH"
};


int main()
{
    /** C Like **/
    cout << "C String Implementation " << ErrorTypeName[COMMAND_OK] << endl;
    
    /** C++ 17 Magic Enum Implementation **/
    cout << "C++ 17 Magic Enum Implementation " << magic_enum::enum_name(ERROR_OUT_OF_BOUNDS);
    return 0;
}
```

Vavat: What if I want to send a speciall message? Do I create an entry in the dictionary for sake of single use? I generally agree that if we had a large somewhat disparate collection of users who where messages might get out of control if left unchecked, proposed solution would be suitable. However, we will only have at most 3-4 controllers, so creating another layer of complexity to solve a problem that'll likely not exist, is poor use of resources. Let's not optimise something that works.

Niuslar (approved):

PR #84, opened 2022-05-10
feature: move MAX_STRING_SIZE definition to etl_profile.h

Niuslar:
There were multiple defines for MAX_STRING_SIZE. Since this is only used with the ETL, I thought it made sense to add this define to etl_profile.h.


MAhmedDev:
In the past kept third party libs clean/upgradeable and moved any system-wide header information to system.h

Vavat (approved):

PR #85, opened 2022-05-11
Feature/nu recursive parser

### LINKED ISSUE #58: Add command processor
Once the UART Com is working, a command processor needs to be added. 
Some options are:
- Writing parser from scratch 
- Use JSON, MessagePack, Protobuf

Niuslar:
I rewrote most of the parser. I tried changing the logic behind the getTokens() method to work as a state-centric state machine, but I didn't like it. I still believe it makes more sense to use the current character as the event and change the status based on this. 

Regarding the actual parser, I'm not sure it is easy to understand, but I don't think any JSON parser is. I added some references so people can understand better the logic behind it.


Niuslar:
If you want to test the parser you could replace the contents of cpp_link.cpp with this: 
```
/**
 * @file cpp_link.cpp
 */

/*
 *  Created on: Feb 11, 2022
 *      Author: niuslar
 */

#include "../etl/to_string.h"
#include "CAdcData.h"
#include "CDebugController.h"
#include "CDispatcher.h"
#include "CJsonParser.h"
#include "CUartCom.h"
#include "adc.h"

/** @note: Instantiating classes here means the ctor for each class will run
 * before any of the setup code is being executed. Therefore, ctor that
 * relies on configured hardware will fail. If class that is meant to run
 * hardware is instantiated, then ctor must only do bare essentials. Once
 * cpp_main() starts executing relevant classes can call initialisation
 * methods that will configure the hardware.
 */
CUartCom g_debug_uart("Main");
CDispatcher g_dispatcher(&g_debug_uart);
CJsonParser g_parser;

/* controllers */
CDebugController g_debug_controller("debug", 100);

#ifdef __cplusplus
extern "C"
{
#endif

    void cpp_main()
    {
        g_debug_uart.init(&huart2);
        g_debug_uart.startRx();

        while (1)
        {
            if (g_debug_uart.isDataAvailable())
            {
                if (g_parser.parse(g_debug_uart.getData()))
                {
                    g_debug_uart.send("Command name: ");
                    g_debug_uart.send(*g_parser.getName());
                    g_debug_uart.send("\n");
                    unsigned int arg_count = g_parser.getArgumentCount();
                    for (unsigned int i = 0; i < arg_count; i++)
                    {
                        g_debug_uart.send("Argument ");
                        etl::string<10> index;
                        etl::to_string((i + 1), index);
                        g_debug_uart.send(index);
                        g_debug_uart.send(": ");
                        etl::string<30> argument;
                        etl::to_string(g_parser[i],
                                       argument,
                                       etl::format_spec().precision(1),
                                       true);
                        g_debug_uart.send(argument);
                        g_debug_uart.send("\n");
                    }
                }
                else
                {
                    g_debug_uart.send("Could not process command\n");
                }
            }
        }
    }

#ifdef __cplusplus
}
#endif
```

Every time you send a command it send back the command name and arguments if the command could be read. Otherwise it send back an error message.

Lib/Inc/CJsonParser.h
```diff
@@ -0,0 +1,74 @@
...
+        CURLY_CLOSE,
+        COLON,
+        STRING,
+        ARRAY_OPEN,
+        ARRAY_CLOSE,
+        COMMA,
+        INVALID
+    } token_type_t;
+
+    typedef struct Stoken
```
Vavat: `SToken` would be more consistent with our naming convention.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+ * checks if the components follow JSON syntax and returns true if it does.
+ */
+/*
+ * Created on : Apr 4, 2022
+ * Author : niuslar
+ */
+
+#include "CJsonParser.h"
+
+#define DEFAULT_VALUE (0.00)
```
Vavat: Two zeros after decimal look a bit weird. Any reason why?

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+        |  Object   | { Pair }     		 |
+        |  Pair     | key : value  		 |
+        |  Key      | string       		 |
+        |  Value    | Number       		 |
+        |  Value    | Array        		 |
+        |  Array    | [ Elements ]       |
+        |  Elements | Number, Number, .. |
+ */
+
+bool CJsonParser::parseObject()
```
Vavat, line 309: Method requires doxygen header.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+                {
+                    b_success = true;
+                }
+            }
+        }
+    }
+    return b_success;
+}
+
+bool CJsonParser::parsePair()
```
Vavat, line 337: Add doxygen header and a description of what this method does.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+        if (m_tokens[m_token_index].type == COLON)
+        {
+            m_token_index++;
+            b_success = parseValue();
+        }
+    }
+    return b_success;
+}
+
+bool CJsonParser::parseValue()
```
Vavat, line 359: Header and a description needed.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+        float number = std::stof(m_tokens[m_token_index].text.c_str());
+        if (m_argument_counter < MAX_ARGUMENT_COUNT)
+        {
+            m_arguments[m_argument_counter++] = number;
+            b_success = true;
+        }
+    }
+    return b_success;
+}
+bool CJsonParser::parseArray()
```
Vavat, line 386: Header and a description needed.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+ * @return True if a JSON object was recognised
+ */
+bool CJsonParser::parse(const etl::string<MAX_STRING_SIZE> &string)
+{
+    // Reset command variables
+    m_argument_counter = 0;
+    m_command_name = "";
+
+    getTokens(string);
+    m_token_index = 0;
```
Vavat: Probably best if this asignment is moved inside the method. Makes much more logical sense inside.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+    }
+    // Finish any token that might have not been stored
+    endToken(current_token);
+}
+
+/**
+ * @brief Add token to m_tokens vector
+ * @param token to be added
+ */
+void CJsonParser::endToken(CJsonParser::token_t &token)
```
Vavat: This name is misleading. This method completes the current token, so perhaps
`completeToken()` or `finishToken()`.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+{
+    if (token.type != WHITESPACE)
+    {
+        if (token.text.empty() == false)
+        {
+            m_tokens.push_back(token);
+            m_token_counter++;
+        }
+    }
+    token.type = WHITESPACE;
```
Vavat: These two lines shouldn't be inside this method. They are essentially cleaning up the token to make it ready for next iteration. So you should put it into a separate method and call it `resetToken()` or `clearToken()`.

Vavat: `resetToken()` is better.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * @brief Access parsed arguments by index of their position in the command
+ * string.
+ *
+ * @param index Position of the argument to be accessed. Zero-based index.
+ * @return Value of the argument at the requested index.
+ */
+float CJsonParser::operator[](unsigned int index)
+{
+    if (index < MAX_ARGUMENT_COUNT)
```
Vavat: Probably not `MAX_ARGUMENT_COUNT`, but `m_argument_counter`, right?

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+            case '1':
+            case '2':
+            case '3':
+            case '4':
+            case '5':
+            case '6':
+            case '7':
+            case '8':
+            case '9':
+                if (current_token.type == STRING ||
```
Vavat: These extended processing inside `case:` statements should be inside separate methods with good descriptive names. Example: `processDigit()`, `processDecimalPoint()`, `processSpacer()`.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+                    endToken(current_token);
+                }
+                break;
+            case '\n':
+            case '\0':
+            case '\r':
+                endToken(current_token);
+                break;
+            case ':':
+                addSpecialChar(current_token, COLON, current_char);
```
Vavat: What's the point of passing `current_token` as argument to a private method? Just creates extra stack operations that are not necessary.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+                addSpecialChar(current_token, ARRAY_OPEN, current_char);
+                break;
+            case ']':
+                addSpecialChar(current_token, ARRAY_CLOSE, current_char);
+                break;
+            case ',':
+                addSpecialChar(current_token, COMMA, current_char);
+                break;
+
+            case '"':
```
Vavat, line 103: `processDoubleQuote()`

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+                    endToken(current_token);
+                    current_token.type = STRING;
+                }
+                else
+                {
+                    endToken(current_token);
+                }
+                break;
+
+            default:
```
Vavat, line 111: Requires explanation when you expect to encounter this branch and what is being done to handle it gracefully.

Lib/Src/CJsonParser.cpp
```diff
@@ -0,0 +1,339 @@
...
+                    current_token.type = INVALID;
+                }
+                else
+                {
+                    current_token.text.append(1, current_char);
+                }
+                break;
+        }
+    }
+    // Finish any token that might have not been stored
```
Vavat, line 115: haha. you even put finish here in the comment, yet the name of the method does not match it's function.

Lib/Src/CJsonParser.cpp
Vavat: Also, there is no need to pass token as argument. Unnecessary stack operation. Instead just call it `finishCurrentToken()`.

Vavat (commented):
Basically a lot of very small, non-functional changes to make code more readable. Main points:

- Names of methods must match exactly what method does.
- Methods must not perform two things, e.g. store current token and prepare next one.
- Unnecessary passing of private members as arguments to private methods. Causes confusion.

Lib/Src/CJsonParser.cpp
Niuslar: This method is outside of the "current_token's" scope, that's why I need to pass it. I prefer this to making another member variable inside the class.

Vavat (approved):

PR #88, opened 2022-05-13
feature: add option for a single string argument in parser

Niuslar:
I added the option to parse a single string as argument (no arrays). I needed this for the Dispatcher and I think it could be useful for other things.

### DIFF
--- a/Lib/Inc/CJsonParser.h
+++ b/Lib/Inc/CJsonParser.h
@@ -26,6 +26,7 @@ public:
     bool parse(const etl::string<MAX_STRING_SIZE> &string);
     const etl::string<MAX_STRING_SIZE> *getName() const;
     unsigned int getArgumentCount() const;
+    const etl::string<MAX_STRING_SIZE> *getStringArgument() const;
     float operator[](unsigned int index);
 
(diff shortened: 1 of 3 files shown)


Lib/Inc/CJsonParser.h
```diff
@@ -26,6 +26,7 @@ class CJsonParser : public ICommand
     bool parse(const etl::string<MAX_STRING_SIZE> &string);
     const etl::string<MAX_STRING_SIZE> *getName() const;
     unsigned int getArgumentCount() const;
+    const etl::string<MAX_STRING_SIZE> *getStringArgument() const;
```
Vavat, line 29: Why are you returning pointer? I thought returning by reference was the preferred way for etl::string. Not criticims, just curious.

Vavat (approved):
This is a true value of good code, when changes like this are easy and straightforward.

Lib/Inc/CJsonParser.h
Niuslar: Just for consistency. There's a getName() method that does almost the same in IComChannel and it returns a pointer. 
I created a ticket the other day to change pointers for references when possible.

PR #90, opened 2022-05-13
fix: re-enable support for floats using sprintf

Niuslar:
For some reason the flags to use floats with sprintf and scanf were not enabled for the Debug build.


Vavat (approved):
That's my fault. I should have done it for both profiles.

PR #91, opened 2022-05-15
Feature/nu dispatcher methods

### LINKED ISSUE #86: Update newCommand() and processComChannels() methods in CDispatcher
We need to update those methods to use the parser

Niuslar:
Fix #86. I added a CJsonParser instance in the Dispatcher class and updated the processComChannels() and newCommand() methods to work with it. I also changed the findControllerNumber() method to return -1 in case no controller was recognised. 

So far the processComChannels() method is fixed with the JSON parser. If we have another parser we can add something like a registerParser() method and then loop through the parsers and see which one recognises the command.


Lib/Src/CDispatcher.cpp
```diff
@@ -133,54 +133,34 @@ void CDispatcher::run()
 /**
  * @brief Method to process incoming commands.
  *
- * @param command Command that arrived via coms channel
+ * @param command ICommand pointer
```
Vavat: The header should add meaning, so it'd be better if this read: Pointer to a command object via an interface class.

Lib/Src/CDispatcher.cpp
```diff
@@ -133,54 +133,34 @@ void CDispatcher::run()
...
-            command.find("(");
-        etl::string<MAX_STRING_SIZE>::size_type close_bracket =
-            command.find(")");
-        if ((open_bracket == command.npos) || (close_bracket == command.npos))
-        {
-            p_comchannel->send("Command: is malformatted.\n");
-        }
-        else
+        int8_t controller_number = findControllerNumber(*controller_name);
+        if (controller_number >= 0)
```
Vavat, line 149: Is it worth having another test that the returned number is not higher than actual number of controllers that were registered? It's perhaps paranoid, but if `findControllerNumber()` method was broken, then weird things will happen here. Perhaps a thought needs to be given to whether this occurence is possible.

Lib/Src/CDispatcher.cpp
```diff
@@ -233,15 +222,17 @@ void CDispatcher::processComChannels()
...
-    uint8_t controller = 0;
-    while ((m_controller_names[controller].compare(name) != 0) &&
-           (controller < m_controller_count))
+    int8_t controller = -1;
+    for (uint8_t i = 0; i < m_controller_count; i++)
     {
-        controller++;
+        if (m_controller_names[i].compare(name) == 0)
+        {
+            controller = i;
```
Vavat, line 234: you can break out of the for loop here if controller is found.
```
controller = i;
break;
```
or you can have `while()` loop instead of `for()` loop:
```
while((controller < 0) && (i < controller_count))
    ...
    ++i;
```

Vavat (approved):
This is excellent stuff. Can you feel how easy it is to modify well designed and maintainable code?

Might be worth creating a ticket to enhance the code with ability to have two different parsers on the same com channel.

Lib/Src/CDispatcher.cpp
Vavat: Probably not. Just being paranoid.

Lib/Src/CDispatcher.cpp
Niuslar: Fixed

Lib/Src/CDispatcher.cpp
Niuslar: Added a break;

PR #94, opened 2022-05-16
refactor: minor changes in CJsonParser to avoid code repetition

Niuslar:
@Vavat I made the changes you suggested, thanks. 
I haven't tested this yet.

### DIFF
--- a/Lib/Src/CJsonParser.cpp
+++ b/Lib/Src/CJsonParser.cpp
@@ -140,15 +140,15 @@ void CJsonParser::processDigit(char current_char)
  */
 void CJsonParser::processDecimalPoint(char current_char)
 {
+    m_current_token.text.append(1, current_char);
+
     if (m_current_token.type == INTEGER)
     {
-        m_current_token.text.append(1, current_char);
         m_current_token.type = FLOAT;
     }
     else
     {
         finishCurrentToken();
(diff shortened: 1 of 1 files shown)


Lib/Src/CJsonParser.cpp
```diff
@@ -140,15 +140,15 @@ void CJsonParser::processDigit(char current_char)
  */
 void CJsonParser::processDecimalPoint(char current_char)
 {
+    m_current_token.text.append(1, current_char);
```
Vavat: This one is a bit suspect, no? In one branch of `if()` if token is integer, character is appended then type is changed to float. That makes sense. In the other branch, however, the sequence changes. It used to be
```
finishCurrentToken();
//then new token begins...
m_curent_token.text.append(1, current_char);
m_curent_token.type = INVALID;
```
In old logic decimal point used to be attached to new token, but now it is attached to old token, which is then finished. But the new token is declared invalid. I think this is an error in logic.

Vavat (changes requested):
I think it's only one place where this refactoring actually makes sense. Apologies for confusion I caused.

Niuslar:
I tried reverting one of the changes but I restored the whole file instead, there's no point in having 3 commits for one small change, I'll create another branch.

PR #95, opened 2022-05-20
Feature/srm plain parser class

### LINKED ISSUE #92: Add support for multiple parser in CDispatcher
The CDispatcher class currently works with the JSON parser. If we want to use multiple parsers, we could add a registerParser() method and loop through all parsers in processComChannels() until the command string is recognised.

Vavat:
* New parser for plain commands in the format: >command(1,2,3).
* Add functionality to have several parsers in CDispatcher.
* Minor refactoring and housekeeping.

### DIFF
--- a/Lib/Inc/CDispatcher.h
+++ b/Lib/Inc/CDispatcher.h
@@ -21,6 +21,7 @@
 
 #include "CController.h"
 #include "CJsonParser.h"
+#include "CStringParser.h"
 #include "CUartCom.h"
 #include "IComChannel.h"
 
@@ -39,6 +40,7 @@ private:
     int8_t findControllerNumber(etl::string<MAX_STRING_SIZE> name);
 
     CJsonParser m_json_parser;
+    CStringParser m_string_parser;
     CUartCom *mp_uart_com;
     CController *mp_controllers[MAX_CONTROLLERS];
     etl::string<MAX_STRING_SIZE> m_controller_names[MAX_CONTROLLERS];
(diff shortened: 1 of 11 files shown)


Lib/Inc/CDispatcher.h
```diff
@@ -47,6 +49,8 @@ class CDispatcher
 
     IComChannel *mp_comchannels[MAX_COMCHANNELS];
     uint8_t m_comchannel_count;
+
+    etl::string<MAX_COMMAND_SIZE> m_command_string;
```
Niuslar, line 53: What is the point of using MAX_COMMAND_SIZE instead of MAX_STRING_SIZE? I think it could lead to errors. Someone might think that by changing MAX_COMMAND_SIZE the parser automatically will recognise longer commands, but this depends on MAX_STRING_SIZE.

Lib/Src/CDispatcher.cpp
```diff
@@ -203,16 +215,14 @@ void CDispatcher::processComChannels()
...
                 }
-            }
-            if (!b_command_recognised)
-            {
-                etl::string<MAX_STRING_SIZE> message;
-                message = "COMMAND_NOT_RECOGNISED\n";
-                mp_comchannels[channel]->send(message);
+                if (!b_command_recognised)
+                {
+                    mp_comchannels[channel]->send(COMMAND_NOT_RECOGNISED);
```
Niuslar, line 221: It would be useful to send a short messages like COMMAND_OK, COMMAND_NOT_RECOGNISED, COMMAND_MALFORMAT for testing.

Niuslar (changes requested):
I think we need to run some tests to check if the parser can filter all invalid commands. I can do that, but it would be much easier if we add the three simple messages I mentioned in a comment.

Niuslar (approved):
Thanks

PR #99, opened 2022-05-24
fix (CStringParser): errors in command parsing.

### LINKED ISSUE #98: Fix string parser
After running the tests I noticed some cases where the parser would pass the command to the controllers even when the format should be invalid. 
There's also a case that breaks the whole application. 

Here's a summary of the failed tests:

| Command tested       | Expected Result | Result  | Possible reason                                            |
|-----------------------------|------------------------|----------------------|----------------------------------------------------------|
|>temperature(arg1, arg2, arg3, ... arg20) [20 values]| Malformat          |  Break the app or OK. | There's no argument limit check in the parser and it's writing into m_arguments beyond its limit |
|~~>temperature(asd25)~~| ~~Malformat~~ | ~~Not recognised~~ |  ~~I think the parser takes that as a string argument~~ |
|>temperature(1,2,,,,,3,4,5,6,7,8,9,0) | Malformat | OK. | I think the parser is storing empty values as arguments? |

Also here's a list of all the commands I tried and their expected results, let me know if you want me to add/change any. 

| Command | Expected Result | 
|---------------|------------------------|
|>?temperature() | COMMAND_OK |
|>temperature(25) | COMMAND_OK |
|command_00  |  COMMAND_MALFORMAT|
|>>temperature(asd25) | COMMAND_~~MALFORMAT~~ OK|
|temperature(25) | COMMAND_MALFORMAT|
|>temperature(1,2,3,4,5,6,7,8,9,0) | COMMAND_OK|
|>temperature(1,2,3,4,5,6,7,8,9,0,1,2,3,4,5,6,7,8,9,0) | COMMAND_MALFORMAT|
|>temperature(1,,,2,3,4,5,6,7,8,9,0) |  COMMAND_MALFORMAT|
|>temperature25) | COMMAND_MALFORMAT|
|>temperature(25 |  COMMAND_MALFORMAT|
|>temperature(25 24 21 53) |  COMMAND_MALFORMAT|

Vavat:
Fix #98.

### DIFF
--- a/Lib/Src/CStringParser.cpp
+++ b/Lib/Src/CStringParser.cpp
@@ -177,13 +177,14 @@ bool CStringParser::parseArguments(const etl::string<MAX_COMMAND_SIZE> &string)
     while (arg_delimiter < arg_stop)
     {
         argument_string = string.substr(arg_start, arg_delimiter - arg_start);
-        if (sscanf(argument_string.c_str(),
-                   "%f",
-                   &(m_arguments[m_argument_counter])) == 1)
+        if ((m_argument_counter < MAX_ARGUMENT_COUNT) &&
+            (sscanf(argument_string.c_str(),
+                    "%f",
+                    &(m_arguments[m_argument_counter])) == 1))
         {
             m_argument_counter++;
         }
-        else if (m_string_argument.empty())
+        else if (!argument_string.empty())
         {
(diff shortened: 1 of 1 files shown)


Niuslar (approved):
If you're happy with this let's merge.

PR #113, opened 2022-06-24
Use ghcr image

duduta:
Hi, I updated the workflow so that it now uses the ghcr image when building the binary.

On an unrelated note, I noticed that the deploy steps, flash_debug and flash_release, are both deploying on hardware (the same hardware?), potentially at the same time. 
Steps in workflows are running in different processes, which could run concurrently on a Raspberry, so it shouldn't work or, at best, it might "work" with unexpected crashes. 
What am I missing?

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -32,15 +32,15 @@ jobs:
         run: |
           chmod +x /opt/st/stm32cubeide_1.9.0/headless-build.sh
           cd /opt/st/stm32cubeide_1.9.0
-          ./headless-build.sh -data /workspace -import /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control -cleanBuild 'Multichannel temperature sensor' -markerType cdt || { echo "Build Failed"; exit 1; }
+          ./headless-build.sh -data /workspace -import /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }} -cleanBuild 'Multichannel temperature sensor' -markerType cdt || { echo "Build Failed"; exit 1; }
       - uses: actions/upload-artifact@v3
         with:
           name: debug_build
-          path: /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Debug/Multichannel\ temperature\ sensor.bin
+          path: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}/Debug/Multichannel\ temperature\ sensor.bin
       - uses: actions/upload-artifact@v3
         with:
           name: release_build
-          path: /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control/Release/Multichannel\ temperature\ sensor.bin
+          path: /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }}/Release/Multichannel\ temperature\ sensor.bin
   
   flash_debug:
     runs-on: self-hosted
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -21,7 +21,11 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    container: niuslar/stm32-cubeide-headless:latest
+    container: 
+      image: ghcr.io/niuslar/stm32-cubeide-headless:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.github_token }}
     steps:
       - uses: actions/checkout@v3
       - name: Build App
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -21,11 +21,7 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    container: 
-      image: ghcr.io/niuslar/stm32-cubeide-headless:latest
-      credentials:
-        username: ${{ github.actor }}
-        password: ${{ secrets.github_token }}
+    container: ghcr.io/niuslar/stm32-cubeide-headless:latest
     steps:
       - uses: actions/checkout@v3
       - name: Build App


Vavat:
> Hi, I updated the workflow so that it now uses the ghcr image when building the binary.
> 
> On an unrelated note, I noticed that the deploy steps, flash_debug and flash_release, are both deploying on hardware (the same hardware?), potentially at the same time. Steps in workflows are running in different processes, which could run concurrently on a Raspberry, so it shouldn't work or, at best, it might "work" with unexpected crashes. What am I missing?
@duduta, You might be right about this not crashing is just us being lucky. The hardware this runs on sits in my office. It's just a rapsberry Pi + nucleo. No hardware attached. We are waiting for the PCB to be manufactured and then we'll have full hardware model of our incubator.
The deploy step does not need to flash both debug and release. In fact I'd argue this is completely unnecessary. Flashing and then thoroughly testing release build should be sufficient. Please, feel free to branch, fix, and submit a PR. 
As I mentioned before we do not have an all seeing management, we try to be self-organising, so everyone throws an idea in the air and if there are not objections we implement it. 
Except perhaps for myself. I only do what's necessary. @FrozenOxide and the rest wouldn't even let me redesign the elevator and pipetter robot so it's prettier. <-- That's an inside joke, btw. I am responsible for hardware and electronics. No time to be fancy. ;-)

Niuslar:
@duduta, thanks for changing the workflow. You're right, I didn't add the condition that flash_debug should finish before running flash_release, but I think on a self-hosted runner only one job can run at a time. 
As @Vavat said, maybe flashing debug and release is unnecessary, feel free to change the code to flash release only.

.github/workflows/c-cpp.yml
```diff
@@ -21,22 +21,26 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    container: niuslar/stm32-cubeide-headless:latest
+    container: 
+      image: ghcr.io/niuslar/stm32-cubeide-headless:latest
```
Niuslar, line 25: In the workflow to build and push the image manually, the image name will vary depending on the username. What happens if someone else runs that workflow?

.github/workflows/c-cpp.yml
```diff
@@ -21,22 +21,26 @@ jobs:
...
+      credentials:
```
Niuslar: Why do we need credentials to run this?

.github/workflows/c-cpp.yml
```diff
@@ -21,22 +21,26 @@ jobs:
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.github_token }}
     steps:
       - uses: actions/checkout@v3
       - name: Build App
         run: |
           chmod +x /opt/st/stm32cubeide_1.9.0/headless-build.sh
           cd /opt/st/stm32cubeide_1.9.0
-          ./headless-build.sh -data /workspace -import /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control -cleanBuild 'Multichannel temperature sensor' -markerType cdt || { echo "Build Failed"; exit 1; }
+          ./headless-build.sh -data /workspace -import /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }} -cleanBuild 'Multichannel temperature sensor' -markerType cdt || { echo "Build Failed"; exit 1; }
```
Niuslar, line 31: Thanks this is great.

duduta:
> 



> > Hi, I updated the workflow so that it now uses the ghcr image when building the binary.
> > On an unrelated note, I noticed that the deploy steps, flash_debug and flash_release, are both deploying on hardware (the same hardware?), potentially at the same time. Steps in workflows are running in different processes, which could run concurrently on a Raspberry, so it shouldn't work or, at best, it might "work" with unexpected crashes. What am I missing?
> > @duduta, You might be right about this not crashing is just us being lucky. The hardware this runs on sits in my office. It's just a rapsberry Pi + nucleo. No hardware attached. We are waiting for the PCB to be manufactured and then we'll have full hardware model of our incubator.
> > The deploy step does not need to flash both debug and release. In fact I'd argue this is completely unnecessary. Flashing and then thoroughly testing release build should be sufficient. Please, feel free to branch, fix, and submit a PR.
> > As I mentioned before we do not have an all seeing management, we try to be self-organising, so everyone throws an idea in the air and if there are not objections we implement it.
> > Except perhaps for myself. I only do what's necessary. @FrozenOxide and the rest wouldn't even let me redesign the elevator and pipetter robot so it's prettier. <-- That's an inside joke, btw. I am responsible for hardware and electronics. No time to be fancy. ;-)
@Vavat Oki doky, I'll open a PR about that. I'm relieved to find out there's no other hardware (other than the Raspberry and Nucleo) attached yet, I've been terrified not to mess up anything. Given that I am null at embedded and engineering. "Elevator" ?? "Pipette"?? . I had to google those, that should tell you about my level, haha.

duduta:
> @duduta, thanks for changing the workflow. You're right, I didn't add the condition that flash_debug should finish before running flash_release, but I think on a self-hosted runner only one job can run at a time. As @Vavat said, maybe flashing debug and release is unnecessary, feel free to change the code to flash release only.

@Niuslar You're welcome, it's my pleasure, I'm grateful that I can contribute. As for the self-hosted runner, I think that the possibility for a race condition is the first thing I will test as soon as I set my Raspberry, it's a yummy subject for me.

.github/workflows/c-cpp.yml
duduta: Hmm, it seems that we don't. I'll remove those. It even works without credentials when the image is private, I tried on another repo. 
I also gotta fix some other thing, I realized that over the weekend, it will take a day or two.

duduta: Hi, I removed the credentials, ready to be merged when you are. (I will open another PR for the other pipeline related stuff that I wanted to polish)

Niuslar (approved):

MAhmedDev (approved):

PR #116, opened 2022-07-01
Inner loop

duduta:
Hi,
This is an attempt to better separate the inner loop from the outer-loop, inspired by this article (it is about DockerHub registry, but I think it applies to GitHub registry as well):
https://www.docker.com/blog/best-practices-for-using-docker-hub-for-ci-cd/

Inner-loop:
- build local docker image
- build app from local image

Outer-loop:
taken care of by the workflows in github.

In case we want to use the official latest ghcr image to test the latest
binary (both debug and release, present in ./Debug and ./Release folders)
- pull latest image from ghcr
- build app from latest image

Some questions remain:
1. do we still need to push/pull to/from DockerHub? Could that code be deleted?
2. does anybody from the team use "make all"? I was thinking to define it to do the inner-loop plus doxy-doc. Is that ok?


Niuslar:
Sorry I thought I left a comment the other day. I was away and didn't have time to look at this, I'll try to review it this evening.

duduta:
It's all right, it was weekend after all :) No hurry, I'm working locally and don't depend on this merge, so take your time. Hope those changes help.

MAhmedDev:
Hi @duduta 
we do not need to keep the docker hub uri if we have github registry working. We can use that for everything. docker-compose-local.yml can be deleted. 

In the future rather than duplicating the docker-compose.yml files with minor changes you can override the base file by using: docker-compose -f file_name_1.yml -f file_name_2.yml build

MAhmedDev:
**do we still need to push/pull to/from DockerHub? Could that code be deleted?**
It can be deleted

**does anybody from the team use "make all"? I was thinking to define it to do the inner-loop plus doxy-doc. Is that ok?**
make all sounds good, feel free to extend the Makefile.

Makefile
```diff
@@ -2,15 +2,21 @@ help: ## make [option]
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
 build-image: ## Build docker CI image locally 
-	DOCKER_BUILDKIT=1 docker-compose -f docker-compose.yml build 
+	DOCKER_BUILDKIT=1 docker-compose -f docker-compose-local.yml build
```
MAhmedDev: you can use the --pull option to pull the latest image from the registry.

docker-compose-local.yml
```diff
@@ -0,0 +1,17 @@
+# declarations to build/run local image used for local development and testing
```
MAhmedDev: I would delete this file as we have moved to github registry

docker-compose-ghcr.yml
```diff
@@ -0,0 +1,18 @@
+# declarations used to pull from ghcr the official latest image used in production
```
MAhmedDev: this can be renamed back to docker-compose.yml

.github/workflows/c-cpp.yml
```diff
@@ -21,22 +21,22 @@ jobs:
...
-    container: niuslar/stm32-cubeide-headless:latest
+    container: ghcr.io/niuslar/stm32-cubeide-headless:latest
     steps:
       - uses: actions/checkout@v3
       - name: Build App
         run: |
           chmod +x /opt/st/stm32cubeide_1.9.0/headless-build.sh
           cd /opt/st/stm32cubeide_1.9.0
-          ./headless-build.sh -data /workspace -import /__w/Multichannel-Temperature-Control/Multichannel-Temperature-Control -cleanBuild 'Multichannel temperature sensor' -markerType cdt || { echo "Build Failed"; exit 1; }
+          ./headless-build.sh -data /workspace -import /__w/${{ github.event.repository.name }}/${{ github.event.repository.name }} -cleanBuild 'Multichannel temperature sensor' -markerType cdt || { echo "Build Failed"; exit 1; }
```
MAhmedDev: good!

docker-compose-ghcr.yml
duduta: ok, I will do that

docker-compose-local.yml
duduta: I will delete all that uses DockerHub.

Makefile
duduta: Hmm, I kinda need to still be able to build the image locally. For example, when trying to integrate Valgrind, I modified the Dockerfile, and I did that locally. I would edit the Dockerfile, build the image locally, test and repeat (what I call the "inner loop"). I did not want nor need to go all the way and push the image to the registry, it would have taken me too long to push and pull at every edit.

MAhmedDev: Hi duduta since the build option is defined in the docker-compose file it will always default to building it locally. If build option did not exist then the build command will pull from the registry (docker-compose -f filename.yml build). The --pull only works when there is no build option defined in the docker-compose, where it will pull the latest every time docker-compose -f filename.yml build --pull is called. 

Run some quick tests and I did not see a difference with the --pull tag when build option is defined in the docker-compose. I would have expected it to have pulled from the registry and then run the local build but that is not the case. The documentation is vague on the details/specifics.

duduta: Oh, sorry, my bad. I thought you meant that I should pull the image from the registry instead of building it, I saw "pull" and I panicked  :)
Yes, you are right, the documentation is vague about the --pull option of the build command. I also did some quick tests, it seems that, when you use the --pull option, it will check all the base images, the "FROM"s (in our case, ubuntu:20.04) to see if what we have locally is the freshest and if it's not, it will fetch it; if not, it will use the ubuntu:20.04 base image that we already have. (It compares the local and remote sha256 hashes). Interesting fact: if you're offline and use the --pull option, it will complain, even if you already have the FROM on your comp. This --pull option seems to be like a guarantee that the image you want to build will be built on the freshest layers possible. Please tell me if you discover otherwise :)

duduta:
Hi, @FrozenOxide, I made the changes you suggested. Feel free to tell me if there's anything else I should do. By the way, thanks for pointing out that I could override the docker-compose.yml file, I didn't know about that, I find it quite elegant.

MAhmedDev:
Hi @duduta docker-compose.local.yml can be deleted since we will likely not use/update the docker hub version.

duduta:
@FrozenOxide Oh, I only focused on changing the tag from "latest" to "local" and forgot to also edit its name and that created confusion.  
This file is only needed in order to change that tag. It's not using DockerHub registry, it is only used for building/running the image locally, for the normal course of development. 
I chose to change the tag in order to see clearly, when using the "docker images" command, where did that image come from: was it built locally (my work in progress) or was it pulled from ghcr. Also, in order to avoid having to rebuild local/ pull latest every time I need to run either of them and the current image under the name ghcr.io/niuslar/stm-32-cubeide-headless:latest is not the one I need to run.

MAhmedDev:
@duduta  docker-compose.local.yml is redundant and can be deleted

duduta:
@FrozenOxide ok, I did that

MAhmedDev (approved):

PR #117, opened 2022-07-20
humidity sensor SPI data readout

samkno1:
Reading humidity value via SPI communication interface for Bosch BME280 sensor.

### DIFF
--- a/Core/Src/main.c
+++ b/Core/Src/main.c
@@ -27,7 +27,7 @@
 
 /* Private includes ----------------------------------------------------------*/
 /* USER CODE BEGIN Includes */
-
+#include <string.h>
 /* USER CODE END Includes */
 
 /* Private typedef -----------------------------------------------------------*/
@@ -47,7 +47,21 @@
 /* Private variables ---------------------------------------------------------*/
 
 /* USER CODE BEGIN PV */
-
+const uint8_t ctrl_meas = 0b01110100;
+const uint8_t ctrl_hum = 0b01110010;
+const uint8_t config = 0b01110101;
+const uint8_t status = 0b11110011;
+const uint8_t hum_msb = 0b11111101;
+const uint8_t temp_msb = 0b11111010;
+const uint8_t dig_H1_addr = 0b10100001;
+const uint8_t dig_H2_addr = 0b11100001;
+const uint8_t dig_H3_addr = 0b11100011;
+const uint8_t dig_H4_addr = 0b11100100;
+const uint8_t dig_H5_addr = 0b11100101;
+const uint8_t dig_H6_addr = 0b11100111;
+const uint8_t dig_T1_addr = 0b10001000;
+const uint8_t dig_T2_addr = 0b10001010;
+const uint8_t dig_T3_addr = 0b10001100;
 /* USER CODE END PV */
 
 /* Private function prototypes -----------------------------------------------*/
@@ -82,7 +96,24 @@ int main(void)
      * 7. Full regression test as much as possible. If don't know what
      * regression test is, read up on it.
      */
-
+	HAL_StatusTypeDef ret;
+	uint8_t hum_val[3]; //humidity raw value
+	uint8_t temp_val[3]; //temperature raw value
+	typedef int32_t BME280_S32_t;
+	typedef uint32_t BME280_U32_t;
+	BME280_S32_t t_fine, adc_T, adc_H;
+	uint8_t spi_buff[5]; //SPI buffer not for measurements
+	uint8_t hum_osrs = 0b00000001; //enable humidity with x1 oversampling
+	uint8_t meas_val = 0b00100001; //set forced mode and turn off press
+	uint8_t filter_off = 0b00000000; //turn off filter
+	BME280_U32_t humidity; //final humidity value
+	uint8_t digT1[2];
+	int8_t digT2[2], digT3[2], digH2[2];
+	unsigned short dig_T1;
+	short dig_T2, dig_T3, dig_H2, dig_H4, dig_H5, dig_H4_ls[2], dig_H5_ls[2];
+	unsigned char dig_H1, dig_H3, digH6;
+	char dig_H6;
+	memset(spi_buff, 0, sizeof(spi_buff));
     /* USER CODE END 1 */
 
     /* MCU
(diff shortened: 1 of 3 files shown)


Core/Src/main.c
```diff
@@ -47,7 +47,21 @@
 /* Private variables ---------------------------------------------------------*/
 
 /* USER CODE BEGIN PV */
-
+const uint8_t ctrl_meas = 0b01110100;
```
Niuslar, line 50: I think this would be much easier to read as hex rather than binary. 
And I would add all of this to a header file.

Core/Src/main.c
```diff
@@ -27,7 +27,7 @@
 
 /* Private includes ----------------------------------------------------------*/
```
Niuslar, line 28: I think working with modules gives us more flexibility, and as you can see at the bottom of the page, main.c is just a link to cpp_main(), which is our main function with C++ support. 

In the Lib folder you can find all the code for the application, I would suggest creating 2 files (.c and .h) just for this sensor. For example (bme280.h and bme280.c)

Core/Src/main.c
```diff
@@ -82,7 +96,24 @@ int main(void)
      * 7. Full regression test as much as possible. If don't know what
      * regression test is, read up on it.
      */
-
+	HAL_StatusTypeDef ret;
+	uint8_t hum_val[3]; //humidity raw value
```
Niuslar, line 100: Perhaps it is easier to just have a longer variable name instead of a comment explaining what it is. 
humidity_raw_value is not much longer and it makes it clearer for future references as well.

Core/Src/main.c
```diff
@@ -82,7 +96,24 @@ int main(void)
...
      */
-
+	HAL_StatusTypeDef ret;
+	uint8_t hum_val[3]; //humidity raw value
+	uint8_t temp_val[3]; //temperature raw value
+	typedef int32_t BME280_S32_t;
+	typedef uint32_t BME280_U32_t;
+	BME280_S32_t t_fine, adc_T, adc_H;
+	uint8_t spi_buff[5]; //SPI buffer not for measurements
+	uint8_t hum_osrs = 0b00000001; //enable humidity with x1 oversampling
```
Niuslar, line 106: I would avoid using "magic numbers". In the header file there could be a list of `#define` with all the constant values. For example: 
```
#define ENABLE_HUMIDITY_X1 0x01
#define TURN_FILTER_OFF 0x00
```

Core/Src/main.c
```diff
@@ -82,7 +96,24 @@ int main(void)
...
+	uint8_t meas_val = 0b00100001; //set forced mode and turn off press
+	uint8_t filter_off = 0b00000000; //turn off filter
+	BME280_U32_t humidity; //final humidity value
+	uint8_t digT1[2];
+	int8_t digT2[2], digT3[2], digH2[2];
```
Niuslar, line 111: I would use `#define` for array sizes as well. Why is it 2?
And I don't understand why you're mixing formats when naming variables. 
filter_off, BME280_U32_t, digT2.
What we've been using is snake_case for variables and camelCase for functions.

Core/Src/main.c
```diff
@@ -117,7 +148,147 @@ int main(void)
...
+    HAL_SPI_Transmit(&hspi2, (uint8_t *)&ctrl_meas, 1, 100);
+    HAL_SPI_Transmit(&hspi2, (uint8_t *)&meas_val, 1, 100);
+    HAL_SPI_Transmit(&hspi2, (uint8_t *)&config, 1, 100);
+    HAL_SPI_Transmit(&hspi2, (uint8_t *)&filter_off, 1, 100);
+    HAL_GPIO_WritePin(GPIOB, GPIO_PIN_12, GPIO_PIN_SET);
+
+    spi_buff[0]=1;
+
+    //check status
+    while(spi_buff[0] & 0b00001000){ //bit masking need bit 3
```
Niuslar, line 167: Here you could use bit shifting to make it more readable. 
If you need bit 3 then `while( spi_buf[0] & ( 1<< 3 ) )` would achieve the same result.
Or you could shift the first argument and compare it to 1. 
`while( (spi_buf[0] >> 3) & 1 )`

Core/Src/main.c
```diff
@@ -82,7 +96,24 @@ int main(void)
      * 7. Full regression test as much as possible. If don't know what
      * regression test is, read up on it.
      */
-
+	HAL_StatusTypeDef ret;
+	uint8_t hum_val[3]; //humidity raw value
+	uint8_t temp_val[3]; //temperature raw value
+	typedef int32_t BME280_S32_t;
+	typedef uint32_t BME280_U32_t;
```
Niuslar, line 103: Why rename uint32_t?

Core/Src/main.c
```diff
@@ -117,7 +148,147 @@ int main(void)
...
+		ret = HAL_SPI_Receive(&hspi2, (unsigned char *)&digH6, 1, 100);
+	}
+
+	dig_H6 = (char) digH6;
+
+	HAL_GPIO_WritePin(GPIOB, GPIO_PIN_12, GPIO_PIN_SET);
+
+	//perform calculations
+	BME280_S32_t var1, var2;
+	var1 = ((((adc_T>>3) - ((BME280_S32_t)dig_T1<<1))) * ((BME280_S32_t)dig_T2)) >> 11;
```
Niuslar, line 277: I think encapsulating this part of the code into functions would be much better. 
If I want a function to read the humidity, I would suggest something like: 
```
float readHumidity()
{ 
    // Code to get humidity
}
```

In the header file you add the function prototype and in the source file you add the function definition.

Niuslar (commented):
Thanks @samkno1, I left some comments with what I think could be improved.

Vavat:
@samkno1, the first thing you should do is turn your code into a module. bme280.h and bme280.c. If you like I can help you with that. This structure is very helpful in exposing just the API access to your module, while hiding anything that is not meant to be used by external entity.

Core/Src/main.c
Vavat: That's actually very common in modular libraries. It somewhat helps with portability if local type is aliased. It would allow for easier changes when code is moved to a different architecture where perhaps 32bit uint is not supported or behaves differently.

Core/Src/main.c
Vavat: @samkno1, I am going to create what Nico is suggesting and push the branch. This way you'll see a clear example of how modules are created. Once architectural problem is solved we can continue the discussion of a more coding style type comments.

Vavat:
https://github.com/Niuslar/Multichannel-Temperature-Control/tree/feature/srm_bme280_refactor

Vavat:
> https://github.com/Niuslar/Multichannel-Temperature-Control/tree/feature/srm_bme280_refactor

@samkno1, I am going to try and refactor your code the way I see fit. If you'd like to follow, feel free. However, I recommend that you have a go yourself. If you want we can have a pair programming session sometime next week. I am back from holiday on 1st of August.

Core/Src/main.c
samkno1: > > https://github.com/Niuslar/Multichannel-Temperature-Control/tree/feature/srm_bme280_refactor
> 
> @samkno1, I am going to try and refactor your code the way I see fit. If you'd like to follow, feel free. However, I recommend that you have a go yourself. If you want we can have a pair programming session sometime next week. I am back from holiday on 1st of August.

Yes, I have already begun restructuring the code. Yes, I'd appreciate that and I am available 6 PM, August 5th (Atlantic Standard Time).

samkno1:
Issue resolved by PR: #119

PR #123, opened 2022-10-13
Feature/srm mains power control

Vavat:
@samkno1, @Niuslar, please review the PR at your leisure. 
There are several things for you to consider when reviewing. 
* This PR does not have an open ticket, therefore, ideally you want to insist on thorough explanation of what is happening, so you can match whether authors intentions were implemented correctly. Author should not force you to guess what PR is trying to achieve.
* This PR is a collection of a small feature + a lot of small refactoring jobs. This is normally frowned upon. One could argue that a better approach would have been to postpone the feature implementation until refactoring of hardware maps and Controller base class was done, then implement a feature on top. This would have resulted in 3 PRs instead of 1 large. The opposing view would be that refactoring without context of a new feature would have been confusing. See if you can formulate an opinion whether you see merits in both approaches. We can discuss it after.

Purpose of PR:
* Implement mains control PWM hardware map methods.
* Implement low level access API to control mains PWM.
* As a result of the above certain refactoring had to be done to CController, IHardwareMap, and two implementations of the hardware map virtual class.


Controllers/CDebugController.cpp
```diff
@@ -10,11 +10,13 @@
...
 #include "CDebugController.h"
+#include "../etl/to_string.h"
 
-CDebugController::CDebugController(etl::string<MAX_STRING_SIZE> name,
+CDebugController::CDebugController(IHardwareMap *p_hardware_map,
+                                   etl::string<MAX_STRING_SIZE> name,
                                    uint32_t run_period_ms)
     : CController(name, run_period_ms),
-      m_breather_light(BREATHING_GPIO_Port, BREATHING_Pin)
+      mp_hw(p_hardware_map)
```
Niuslar, line 19: I think we should check that this is not a`nullptr` somewhere.

Controllers/CDebugController.cpp
```diff
@@ -26,14 +28,110 @@ CDebugController::~CDebugController()
 
 void CDebugController::run()
 {
-    m_breather_light.toggle();
+    static float breathing_light_intensity = 0;
+    mp_hw->setBreathingLight(breathing_light_intensity);
+    breathing_light_intensity = breathing_light_intensity ? 0 : 100;
```
Niuslar, line 36: If I understood correctly, this function sets the duty cycle to 0 or 100 to turn the LED on/off every time `run()` is called?

Controllers/CDebugController.cpp
```diff
@@ -26,14 +28,110 @@ CDebugController::~CDebugController()
...
+        error_code = getMainsPower(p_command, &mains_power);
+        if (error_code == ICommand::COMMAND_OK)
+        {
+            CController::s_scratch_pad = "Mains channel ";
+            etl::to_string((*p_command)[0],
+                           CController::s_scratch_pad,
+                           etl::format_spec(),
+                           true);
+            CController::s_scratch_pad += ": ";
+            etl::to_string(mains_power,
```
Niuslar, line 66: It would be great to have an overloaded "addToString()" function that adds floats, ints, and string to the buffer before calling send(). I might have a look this weekend.

Controllers/CDebugController.cpp
```diff
@@ -26,14 +28,110 @@ CDebugController::~CDebugController()
...
+    {
+        mains_channel = 0;
+        mains_power = (*p_command)[0];
+    }
+    else if (p_command->getArgumentCount() == 2)
+    {
+        mains_channel = (uint8_t)(*p_command)[0];
+        if (mains_channel != (*p_command)[0])
+        {
+            return (ICommand::ERROR_TYPE_MISMATCH);
```
Niuslar, line 99: Will this ever be true? How can `mains_channel != (*p_command)[0]` if we are assigning it that value the line before?

Hardware/CRealHardwareMap.cpp
```diff
@@ -167,3 +176,102 @@ void CRealHardwareMap::enableControlPower(bool b_enable)
...
+    if (power < 0)
+    {
+        power = 0;
+    }
+    if (power > 100)
+    {
+        power = 100;
+    }
+    uint16_t ccr = p_timer->Instance->ARR * mainsPowerCorrection(power) / 100;
+    p_timer->Instance->CCR1 = ccr;
```
Niuslar, line 208: This doesn't change anything, but the HAL has a `__HAL_TIM_SET_COMPARE(__HANDLE__, __CHANNEL__, __COMPARE__) ` "function" that sets the CCR register.

Controllers/CDebugController.cpp
```diff
@@ -26,14 +28,110 @@ CDebugController::~CDebugController()
...
+        }
+        mains_power = (*p_command)[1];
+    }
+    // range check arguments.
+    if ((mains_channel > 2) || (mains_power < 0) || (mains_power > 100))
```
Niuslar, line 108: I think we need to fix this because if `_command->getArgumentCount()` is not 1 or 2, then mains_channel  and mains_power will be uninitialised.

Niuslar (commented):
There's only one thing I think we might need to check. The rest looks good, thanks.

Controllers/CDebugController.cpp
Vavat: Yes. This is really just a plug for when we get around to beautifying it. Ideally, we would want CDebugController to be able to switch this light from slow flash for "Everything is OK" to other types of flashing for when something is the matter and might require attention. But this is frills and cannot distract us right now.

Controllers/CDebugController.cpp
Vavat: That'd be awesome. Like c# and java do.

Controllers/CDebugController.cpp
Vavat: This is a bit of an awkward way of seeing if an argument is a `float`. Horrible solution, but I couldn't come up with a better one. Open to suggestions.

Hardware/CRealHardwareMap.cpp
Vavat: I dislike macros. Any compelling reason I should use your suggestion?

Controllers/CDebugController.cpp
Vavat: That's because I forgot final `else{}`. Fixed.

Controllers/CDebugController.cpp
Vavat: One option I can think of is to simply cast whatever is in the first argument into unsigned int and then range check. If it was sent as float, then we ignore the decimal part and use it as if it was an integer. This will not cause failures, but it'd be possible to request channel 2.5 to switch to 100% power, which will switch channel 2 or will bounce back with error depending on how casting works.

Controllers/CDebugController.cpp
Vavat: What might be a good thing to do is create a child class based on `etl::string` and expand it with ability to absorb numbers. 
Another option is to expand `send()` method within `CComsController` to be able to accept parameters like `sprintf()` does.

PR #125, opened 2022-12-30
Samk bme280 hardware integration

samkno1:
New feature for humidifier and fix for CBME280 class


Controllers/CHumidityController.cpp
```diff
@@ -0,0 +1,192 @@
...
+ */
+
+#include "CHumidityController.h"
+#include <stdio.h>
+#include "CBME280.h"
+#include "CHumidifier.h"
+#include "ICommand.h"
+#include "IHardwareMap.h"
+
+#define MIN_TEMPERATURE  10
```
Niuslar: I know we have used `#define` macros quite a lot in this code, but I think it is safer to use something like `static constexpr int MIN_TEMPERATURE = 10` inside the class. 
I might refactor the other classes that use macros to use `constexpr` instead.

Controllers/CHumidityController.cpp
```diff
@@ -0,0 +1,192 @@
...
+#define MAX_TEMPERATURE  50
```
Niuslar: I cannot find where some of these macros are being used. Will we need them in the future?

Lib/Inc/CBME280.h
```diff
@@ -0,0 +1,149 @@
...
+ *
+ *  Created on: 24 Jul 2022
+ *      Author: salavat.magazov
+ */
+
+#ifndef CBME280_H_
+#define CBME280_H_
+
+#include "CGpioWrapper.h"
+#include "gpio.h"
```
Niuslar, line 17: I think we could remove this `#include "gpio.h"` as it is included with `CGpioWrapper.h`, but if you want to keep it for clarity I understand.

Controllers/CHumidityController.h
```diff
@@ -0,0 +1,46 @@
...
+#ifndef CHUMIDITYCONTROLLER_H_
+#define CHUMIDITYCONTROLLER_H_
+
+#include "CBME280.h"
+#include "CController.h"
+#include "CHumidifier.h"
+#include "CPIDLoop.h"
+#include "IHardwareMap.h"
+
+#define CHANNEL_NUMBER 1
```
Niuslar: If the scope of this constant is just this class, I think a private member variable would be much better. 
`static constexpr uint8_t CHANNEL_NUMBER = 1`

Lib/Inc/CBME280.h
```diff
@@ -0,0 +1,149 @@
...
+    } m_state;
+
+    static CBME280 *sp_sensors[MAX_SENSORS];
+    static uint8_t s_sensor_count;
+
+    float m_temperature_calibration[T_CALIBRATION_SIZE];
+    float m_pressure_calibration[P_CALIBRATION_SIZE];
+    float m_humidity_calibration[H_CALIBRATION_SIZE];
+
+    uint8_t m_raw_adc_data[RAW_ADC_DATA_SIZE];
```
Niuslar, line 138: I checked where this array is being used and I think the name is a bit confusing. This makes me think we're using ADC channels with this sensor, and I don't think that's the case. Please let me know if I'm wrong.

Lib/Src/CBME280.cpp
```diff
@@ -0,0 +1,393 @@
...
+{
+    if (HAL_SPI_GetState(mp_spi) != HAL_SPI_STATE_READY)
+    {
+        return false;
+    }
+    /* Configure SPI comms as required by the sensor. */
+    // TODO: this type of channel sharing with different
+    // configurations should be handled by a wrapper class. It can
+    // also queue up traffic and call callback functions.
+    mp_spi->Init.CLKPolarity = SPI_POLARITY_HIGH;
```
Niuslar, line 301: What else are we using SPI for?

Niuslar (commented):
@samkno1 I left some comments. This looks very good. I still need to look at the CBMR280 class in more detail before I can make any comments on that.

Controllers/CHumidityController.cpp
samkno1: No, temperature values are not required.

Controllers/CHumidityController.h
samkno1: I agree

Lib/Inc/CBME280.h
samkno1: Yes, we do not use ADC channels to communicate with the sensor. However, the sensor incorporates an internal ADC in which data is transferred from to the SPI registers.

Lib/Src/CBME280.cpp
samkno1: @Vavat

Controllers/CHumidityController.cpp
Vavat: Agreed. @samkno1, there are two ways you can set constants in your code:
`#define CONSTANT_VALUE 10`
or
`constexpr unsigned int CONSTANT_VALUE = 10;`

Fundamentally, they both achieve the same thing, as the values will be resolved at compile time, but `#define` is resolved by pre-processor that has no logic in it. It blindly replaces every instance of `CONSTANT_VALUE` with a number 10. While if you declare it as `const` or `constexpr` then compiler will make sure that types are correct. 
Generally, using `const` or `constexpr` is considered safer and a better practice. As far as memory usage goes, compiler will optimise the values out, so memory usage will not be impacted.
Finally, `const` vs. `constexpr` is a bit subtle. You can use `constexpr` everywhere. The difference was introduced in 2014 edition, IIRC, and allows you to not only designate values, but also expressions. Example:
`constexpr float HALF_ANGLE_SIN = sin(45);`
as long as the expression can be evaluated at compile time and not at runtime, it can be declared as `constexpr`.

Controllers/CHumidityController.cpp
Vavat: In case there is legacy stuff, don't be afraid to remove it. Since we are in full version control, they can always be restored. It keeps the code clean and much easier to follow and develop.

Lib/Src/CBME280.cpp
Vavat: In this particular case we are not going to share SPI. If we ever go towards sharing SPI channel between different controllers, then we'd have to write a separate wrapper that would allow marshalling of multiple access to the same hardware. This is one of the reasons FreeRTOS is an attractive solution. It has built-in mechanisms to achieve that called Semaphore and Mutex. Semaphore allows limiting the number of access token in situations where hardware can service limited number of requests simultaneously. While Mutex (portmanteau of MUTually EXclusive) is a mechanism to allow only one entity to access the hardware. Mutex in FreeRTOS also allows for prioritised queueing.

Vavat (commented):
@samkno1, I forgot how much code you wrote and how complex this is. I am going to have to spend a bit more time looking into it and compiling and trying to run it. If you make small changes like #define and removing unused stuff in the meantime. I'll endeavour to do a full review by end of the week.

Vavat (commented):
@samkno1, I've reviewed this code. Aside from minor stuff, this is massive progress. Have a look at a separate branch I created to do some very minor clean up. As I said, a lot of it is just removing legacy stuff and some is a matter of taste.
I also cleaned up api a little.

samkno1:
> @samkno1, I've reviewed this code. Aside from minor stuff, this is massive progress. Have a look at a separate branch I created to do some very minor clean up. As I said, a lot of it is just removing legacy stuff and some is a matter of taste. I also cleaned up api a little.

I'll take a look!

